PR #2, opened 2020-05-21
Sacn receiver change universe

ChristianReese:
So far this PR encompasses the sacn_receiver_change_universe function, the revised receiver example application, an integration test (still in the unit test folder), and a unit test, alongside minor refactors sprinkled around. I'm posting this PR now even though I realize more unit tests will probably be needed before merging. This way you can have a look at this earlier, and we'll have a platform for discussion. Next I am planning on adding unit tests to test change_universe's error conditions to make sure they get returned as expected. If you think of any other ideas for unit tests please let me know.


examples/receiver/src/main.c
```diff
@@ -94,41 +97,271 @@ static SourceData* find_source_hole(ListeningUniverse* listener)
...
-          int update_rate = source->num_updates * 1000 / interval_ms;
-          printf("  Source %s\tPriority: %u\tUpdates per second: %d\tLast update: ", cid_str, source->priority,
-                 update_rate);
-          for (size_t i = 0; i < NUM_SLOTS_DISPLAYED; ++i)
+          if (source->valid)
           {
-            printf("%02x ", source->last_update[i]);
+            char cid_str[ETCPAL_UUID_STRING_BYTES];
+            etcpal_uuid_to_string(&source->cid, cid_str);
+            int interval_ms = etcpal_getms() - source->update_start_time_ms;
```
samkearney, line 267: interval_ms should be uint32_t

src/sacn/private/common.h
```diff
@@ -109,6 +109,8 @@ typedef unsigned int sacn_thread_id_t;
 #define SACN_CAN_LOG(pri) false
 #endif
 
+#define UNIVERSE_ID_VALID(universe_id) ((universe_id != 0) && (universe_id <= 64000))
```
samkearney, line 112: Nice. Good idea.

ChristianReese:
Moved the integration test into another branch because it was failing to build on Mac and Linux. Oddly enough, GitHub appears to be associating the latest build with the commit that failed, based on where it displays the checkmark.

src/sacn/receiver.c
```diff
@@ -480,6 +546,18 @@ SacnReceiver* create_new_receiver(const SacnReceiverConfig* config)
 
   receiver->socket = ETCPAL_SOCKET_INVALID;
 
+  if (config->netints != NULL)
+  {
+    receiver->netints = calloc(config->num_netints, sizeof(SacnMcastNetintId));
```
samkearney, line 551: I understand the netints now need to be saved, so that change_universe will work. This calloc() is not guarded by an #ifdef SACN_DYNAMIC_MEM right now. You will need to tweak the SacnReceiver struct to have something like:

```c
  /* (optional) array of network interfaces on which to listen to the specified universe. If NULL,
   * all available network interfaces will be used. */
#ifdef SACN_DYNAMIC_MEM
  const SacnMcastNetintId* netints;
#else
  const SacnMcastNetintId netints[SACN_MAX_NETINTS];
#endif
  /* Number of elements in the netints array. */
  size_t num_netints;
```

Then change this code so that it either does the calloc if SACN_DYNAMIC_MEM==1, or simply fills in the static array if SACN_DYNAMIC_MEM==0. It should also fail if the calloc fails, or if too many netints are specified for the static array. This logic is probably worth refactoring into its own small static function to avoid polluting this function with #ifdefs.

tests/unit/receiver/test_receiver.cpp
```diff
@@ -80,3 +82,55 @@ TEST_F(TestReceiver, SetExpiredWaitWorks)
   sacn_receiver_set_expired_wait(std::numeric_limits<uint32_t>::max());
   EXPECT_EQ(sacn_receiver_get_expired_wait(), std::numeric_limits<uint32_t>::max());
 }
+
+TEST_F(TestReceiver, ChangeUniverseWorks)
+{
+#define CHANGE_UNIVERSE_WORKS_FIRST_SOCKET 1u
```
samkearney, line 88: Change these to constexpr uint16_t (nice modern C++ alternative to #defines)

tests/unit/receiver/test_receiver.cpp
```diff
@@ -80,3 +82,55 @@ TEST_F(TestReceiver, SetExpiredWaitWorks)
...
+  };
+
+  etcpal_error_t change_universe_result = sacn_receiver_change_universe(handle, CHANGE_UNIVERSE_WORKS_SECOND_UNIVERSE);
+
+  EXPECT_EQ(change_universe_result, kEtcPalErrOk);
+  EXPECT_EQ(sacn_lock_fake.call_count, sacn_unlock_fake.call_count);
+  EXPECT_EQ(sacn_initialized_fake.call_count, 2u);
+  EXPECT_EQ(clear_term_set_list_fake.call_count, 1u);
+  EXPECT_EQ(sacn_remove_receiver_socket_fake.call_count, 1u);
+  EXPECT_EQ(sacn_add_receiver_socket_fake.call_count, 2u);
```
samkearney, line 129: Awesome.

src/sacn/private/pdu.h
```diff
@@ -70,8 +70,8 @@ bool parse_sacn_data_packet(const uint8_t* buf, size_t buflen, SacnHeaderData* h
                             const uint8_t** pdata);
 bool parse_draft_sacn_data_packet(const uint8_t* buf, size_t buflen, SacnHeaderData* header, uint8_t* seq,
                                   bool* terminated, const uint8_t** pdata);
-void pack_sacn_data_header(uint8_t* buf, const EtcPalUuid* source_cid, const char* source_name, uint8_t priority,
-                           bool preview, uint16_t universe_id, uint8_t start_code, uint16_t slot_count);
+size_t pack_sacn_data_header(uint8_t* buf, const EtcPalUuid* source_cid, const char* source_name, uint8_t priority,
```
samkearney, line 73: This change is fine, but I'm not seeing why it was necessary?

ChristianReese: I believe this was utilized in the integration test, which has moved to a different branch.

tests/unit/receiver/test_receiver.cpp
ChristianReese: Thanks!

src/sacn/private/common.h
ChristianReese: Thanks!

src/sacn/receiver.c
ChristianReese: Will work on this next week.

ChristianReese:
@samkearney I've addressed most of your comments in this commit - I'll need to work on the last one next work week. I also added some empty test definitions to test if the given errors occur under correct circumstances. I figured these would be good to have as tests since it goes along with the documentation. Next time I plan to develop these as well.

src/sacn/receiver.c
ChristianReese: The code for this has just been committed.

ChristianReese:
This commit addresses the dynamic netints array thing. I had to make some adjustments to the memory unit tests - hopefully those are okay, the unit tests are still succeeding. The next commit will probably be the unit tests for change_universe's error conditions.

ChristianReese:
I implemented the unit tests for change_universe's error conditions. Does this PR look like it's ready to pull into develop @samkearney ? Any more unit test ideas you can think of?

samkearney:
This looks good! I say merge.

PR #3, opened 2020-06-04
Fix for C4090 + free safety in FREE_RECEIVER

ChristianReese:
I still have to look into the C4100 warnings - starting this now so that I can look at a new build with the right line numbers.


ChristianReese:
@samkearney good to merge?

PR #5, opened 2020-06-05
C Merger API proposal

nwagner:
Here's a draft of the API I'm proposing for the C DMX Merger API. Let me know what you think!


include/sacn/merge.h
```diff
@@ -0,0 +1,162 @@
...
+/*!
+ * \file sacn/merge.h
+ * \brief DMX Merge API definitions
+ *
+ * Functions and definitions for the \ref dmx_merger "DMX Merger API" are contained in this
+ * header.
+ */
+
+#ifndef DMX_MERGER_H_
+#define DMX_MERGER_H_
```
samkearney, line 29: I've adopted a convention for these projects where the header guard matches the file name (scoped with subdirectories starting with the subdirectory matching the name of the library).

So, this should be:
```
#ifndef SACN_MERGE_H_
#define SACN_MERGE_H_
```

include/sacn/merge.h
```diff
@@ -0,0 +1,162 @@
...
+  /*! Whether or not the address_priority buffer is valid. */
+  bool address_priority_valid;
+
+  /*! The sACN per-address (startcode 0xdd) priority (1-255, 0 means not sourced).
+      If the source does not */
+  uint8_t address_priority[DMX_ADDRESS_COUNT];
+
+} DmxMergerSource;
+
+etcpal_error_t dmx_merger_init();
```
samkearney, line 133: I'm not a huge fan of the name `dmx_merger`. For a few reasons:

* It does not start with `sacn_` and so breaks the convention of the rest of this library and other libraries, that try to always pseudo-namespace the library name to the front of all modules.
* "merger" is a noun, but it doesn't describe the handle/object instance that is created when using this API. For a counterexample, `sacn_receiver_create()` creates an sACN receiver instance, which you access using a handle called `sacn_receiver_t`. But these functions create "universes", which you access using a `universe_handle_t`.

A few alternatives that come to mind:

**sacn_merger**
```c
/*! \file sacn/merger.h */

typedef int sacn_merger_t

etcpal_error_t sacn_merger_create(const SacnMergerConfig* config, sacn_merger_t* handle);
```

**sacn_dmx_merger**
```c
/*! \file sacn/dmx_merger.h */

typedef int sacn_dmx_merger_t;

etcpal_error_t sacn_dmx_merger_create(const SacnDmxMergerConfig* config, sacn_dmx_merger_t* handle);
```

Notice in all cases that I've tried to keep mnemonic consistency between the file name, the handle type, and the API function names as much as possible.

include/sacn/merge.h
```diff
@@ -0,0 +1,162 @@
...
+ * \defgroup dmx_merger DMX Merger
+ * \ingroup sACN
+ * \brief The DMX Merger API
+ *
+ * This API provides a software merger for buffers containing DMX512-A start code 0 packets.
+ * It also uses buffers containing DMX512-A start code 0xdd packets to support per-address priority.
+ *
+ * When asked to calculate the merge of a universe, the merger shall evaluate the current source
+ * buffers and update two result buffers:
+ *   512 bytes for the merged data values (i.e. "winning level").  These are calculated by using
```
samkearney, line 48: If you prefix these lines with another * (or a -, or a +, or I think a few other "bullet" types are valid), Doxygen will convert them to bullets using markdown conventions.

FYI in case you weren't aware, the CI is set up to generate staging docs for each pull request. So you can check what the generated documentation for this PR looks like at the URL matching the number of this pull request:
https://etclabs.github.io/sACN/docs/stage/5

This paragraph is rendered at https://etclabs.github.io/sACN/docs/stage/5/group__dmx__merger.html#details

include/sacn/merge.h
```diff
@@ -0,0 +1,162 @@
...
+
+} DmxMergerSource;
+
+etcpal_error_t dmx_merger_init();
+void dmx_merger_deinit(void);
+
+etcpal_error_t dmx_merger_create_universe(const DmxMergerUniverseConfig* config, universe_handle_t* handle);
+etcpal_error_t dmx_merger_destroy_universe(universe_handle_t handle);
+
+etcpal_error_t dmx_merger_add_source(universe_handle_t universe, const EtcPalUuid* source_cid, source_id_t* source_id);
```
samkearney, line 139: So does this API wrap the sACN Receiver API, or is the user expected to add sources manually using this function when a new source is delivered via the universe_data callback?

include/sacn/common.h
```diff
@@ -49,6 +49,9 @@ extern "C" {
  */
 #define SACN_SOURCE_NAME_MAX_LEN 64
 
+/*! The number of addresses in DMX universe. */
```
samkearney, line 52: Should say in *a* DMX universe

include/sacn/merge.h
nwagner: The main reason I chose dmx_merger (and had it in its own library) was because I wanted to emphasize that this could be used for sACN or in combination with other DMX sources.   But I do see your point.  I'll play with your suggestions.

include/sacn/merge.h
nwagner: It does not wrap the Receiver API.  You're right that there's currently an impedance mismatch.  I am planning on changing the Receiver API to have a SourcesFound callback rather than relying on the UniverseData callback to cause source creation.  Also, I wanted users to be able to add sources for other inputs.

My goal is to have the outputs of the Receiver API be trivial to pipe to this API, so you're right that we'll want to keep these issues in mind.

include/sacn/merge.h
nwagner: Thanks for the heads up on the docs!

include/sacn/merge.h
samkearney: OK, I get it. In that case I think we should have another API, which links them together, for maximum ease of use.

samkearney: Basically I want the all-software sACN merge case (e.g. network fixtures) to be able to use a single API and not have to hook two together.

nwagner: I chose this method so that the application can choose what to do with the received sACN output: e.g. give it to the FPGA or give it to the merger.

The merger currently doesn't do anything asynchronous.  If the merger was able to register callbacks with the receiver library, would it then need an notification so the app knows the values have changed, or would the app free wheel over the buffer?

samkearney: I would imagine that being the job of the yet-unnamed third, encompassing API.

So for example:

```
  sACN Merge Receiver
    /             \
sACN Receiver  DMX Merger
```

DMX Merger would remain all-synchronous.
Apps that use hardware merging would use sACN receiver only.
Apps that merge sACN with other DMX sources would use sACN Receiver and DMX Merger, hooking them up manually.
Apps that only need software-merged sACN could use sACN Merge Receiver.

include/sacn/merge.h
nwagner: Everything should be changed over to sacn_dmx_merger now.  This feels right -- I didn't really like the notion of universe bleeding into this module, and now I was able to remove it.  Let me know if you have other issues or resolve the conversation.

include/sacn/merge.h
nwagner: I see your point, but I hesitate.  Would it be ok for us to see how these libraries go before specifying and reviewing requirements/api for an additional library?  If the API is nearly a pass-through, we could provide examples for the sACN receiver APIs that show the needed call to the merger.

samkearney: Oh, I wasn't thinking of an additional library, just another header/module in this library.

nwagner: Fair enough, I've noticed that I tend to conflate the two.  But it still would mean requirements & review.

samkearney: I guess I don't foresee it being an especially contentious review item? Where is your hesitation coming from?

nwagner: Two mental blocks come to mind, both of which may be completely invalid:
1) This feels like ACNConnect, and that was added to cover app-unfriendliness in the lower library.  I'm worried that we're adding unnecessary layers.
2) I also have a feeling that applications may want to use this data differently (e.g. a notification on every change vs. freewheeling on the buffers).

src/sacn/dmx_merger.c
```diff
@@ -0,0 +1,314 @@
...
+ *
+ * Creates a new merger that uses the passed in config data.  The application owns all buffers
+ * in the config, so be sure to call dmx_merger_destroy before destroying the buffers.
+ *
+ * \param[in] config Configuration parameters for the DMX merger to be created.
+ * \param[out] handle Filled in on success with a handle to the merger.
+ * \return #kEtcPalErrOk: Merger created successful.
+ * \return #kEtcPalErrInvalid: Invalid parameter provided.
+ * \return #kEtcPalErrNotInit: Module not initialized.
+ * \return #kEtcPalErrNoMem: No room to allocate memory for this receiver.
```
ChristianReese, line 143: for this _merger_

src/sacn/dmx_merger.c
```diff
@@ -0,0 +1,314 @@
...
+ * \return #kEtcPalErrSys: An internal library or system call error occurred.
+ */
+etcpal_error_t sacn_dmx_merger_create(const SacnDmxMergerConfig* config, sacn_dmx_merger_t* handle)
+{
+  return kEtcPalErrNotImpl;
+}
+
+/*!
+ * \brief Destroy an merger instance.
```
ChristianReese, line 152: _a_ merger instance

src/sacn/dmx_merger.c
```diff
@@ -0,0 +1,314 @@
...
+ *   - It is the source identifer that is put into the slot_owners buffer that was passed
+ *     in the DmxMergerUniverseConfig structure when creating the merger.
+ *
+ * \param[in] merger The handle to the merger.
+ * \param[in] source_cid The sACN CID of the source.
+ * \param[out] source_id Filled in on success with the source id.
+ * \return #kEtcPalErrOk: Source added successfully.
+ * \return #kEtcPalErrInvalid: Invalid parameter provided.
+ * \return #kEtcPalErrNotInit: Module not initialized.
+ * \return #kEtcPalErrNoMem: No room to allocate memory for this receiver, or the max number of sources has been
```
ChristianReese, line 182: for this _source_

src/sacn/dmx_merger.c
```diff
@@ -0,0 +1,314 @@
...
+
+/*!
+ * \brief Gets a read-only view of the source data.
+ *
+ * Looks up the source data and returns a pointer to the data or NULL if it doesn't exist.
+ * This pointer is owned by the library, and must not be modified by the application.
+ * The pointer will only be valid until the source or merger is removed.
+ *
+ * \param[in] merger The handle to the merger.
+ * \param[in] source The id of the source to remove.
```
ChristianReese, line 218: Remove "to remove"

src/sacn/dmx_merger.c
```diff
@@ -0,0 +1,314 @@
...
+ * \return #kEtcPalErrOk: Source updated and merge completed.
+ * \return #kEtcPalErrInvalid: Invalid parameter provided.
+ * \return #kEtcPalErrNotInit: Module not initialized.
+ * \return #kEtcPalErrNotFound: Handle does not correspond to a valid source or merger.
+ * \return #kEtcPalErrSys: An internal library or system call error occurred.
+ */
+// TODO: If Receiver API changes to notify both values and per-address priority data in the same callback, this should
+// change!!
+etcpal_error_t sacn_dmx_merger_update_source_from_sacn(sacn_dmx_merger_t merger, source_id_t source,
+                                                       const SacnHeaderData* header, const uint8_t* pdata)
```
ChristianReese, line 275: If I'm calling this from SacnUniverseDataCallback then I'm going to have to track the source_id_t since the callback will only give me the CID. Unless if source_id_t could become a part of the receiver API as well?

src/sacn/dmx_merger.c
```diff
@@ -0,0 +1,314 @@
...
+  return kEtcPalErrNotImpl;
+}
+
+/*!
+ * \brief Fully recalculate outputs.
+ *
+ * Does a full recalculation of the merger outputs.
+ *
+ * \param[in] merger The handle to the merger.
+ * \return #kEtcPalErrOk: Source updated and merge completed.
```
ChristianReese, line 305: I'd just say "recalculation completed."

include/sacn/dmx_merger.h
```diff
@@ -0,0 +1,166 @@
...
+                                                  const uint8_t* new_values, size_t new_values_count, uint8_t priority,
+                                                  const uint8_t* address_priorities, size_t address_priorities_count);
+// TODO: If Receiver API changes to notify both values and per-address priority data in the same callback, this should
+// change!!
+etcpal_error_t sacn_dmx_merger_update_source_from_sacn(sacn_dmx_merger_t merger, source_id_t source,
+                                                       const SacnHeaderData* header, const uint8_t* pdata);
+etcpal_error_t sacn_dmx_merger_stop_source_per_address_priority(sacn_dmx_merger_t merger, source_id_t source);
+
+// TODO: Do we need this?
+etcpal_error_t sacn_dmx_merger_recalculate(sacn_dmx_merger_t merger);
```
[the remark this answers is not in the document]
nwagner: And not necessarily even then.  The update functions could be optimized to only need to calculate changes on a particular source, and only look at the other sources when that source has changed priority or given up control of the channels.  I seem to recall Dan's merger doing it this way.

This function would be a complete recalc.

src/sacn/dmx_merger.c
nwagner: That was an oversight, Good Catch!  I've added a get_id function for general use, and removed the id from the parameters of this function.

include/sacn/dmx_merger.h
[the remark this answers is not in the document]
ChristianReese: It sounds like we could go either way with this. We could either keep it in case anyone finds a use for it, or get rid of it and see if anyone brings up a need for it in a larger review.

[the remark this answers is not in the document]
nwagner: Let's keep it in until you've finished your rough implementation.  If the other functions don't use it, it would be simpler to drop this function (unless the review wants it, as you say).

nwagner:
Ok, that's all the review we needed for this round.   Thanks everyone!

PR #6, opened 2020-08-05
Offensive terminology refactor

ChristianReese:


docs/Doxyfile
```diff
@@ -1345,7 +1345,7 @@ CHM_FILE               =
 HHC_LOCATION           = 
 
 # The GENERATE_CHI flag controls if a separate .chi index file is generated
-# (YES) or that it should be included in the master .chm file (NO).
+# (YES) or that it should be included in the main .chm file (NO).
```
samkearney, line 1348: This file is generated by Doxygen (we didn't write it) and will be replaced wholesale when we update our Doxygen version. This should be reverted.

PR #9, opened 2020-10-12
Initial ratified transmit api

nwagner:
This is my initial proposal for the ratified sACN source API.  Above and beyond this review, there are a couple of issues to think about.
1) Because draft is not transmitted by this library, we don't control draft source sequence numbers.  Do we need to worry about an ETC device sourcing both draft & ratified for the same universe?  I really don't want to have to put filtering in the receive side for this case, since it should be rare.  I'm going to ping Ray on this one, too.
2) I'm having difficulty thinking of an abstract interface class for the sACN source that would provide something like Eos is looking for and still be useful.  Or even make sense to be exposed in open source.  Maybe that group would need to create their own layer for their own needs, or this is something additional we could provide.  Again, I'll ping Ray.


nwagner:
Oops -- the github desktop tried to create this pull request from develop instead of api_proposal.

PR #11, opened 2020-10-27
Universe discovery

nwagner:
Here's my proposal for a very slim Universe Discovery api.  As part of doing this, I realized that we didn't have the test frameworks in yet for a few items, so I added those as well.  Christian, if this steps on your toes we can work it out.

After this, we should look into any API or comment/documentation changes needed from Christian's current work, merge them, and start the general review! If Dan gets some suggestions to us first, we can incorporate them as well.


include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,327 @@
...
+
+    /**
+     * @brief Notify that the module has run out of memory to track universes or sources
+     *
+     * If #SACN_DYNAMIC_MEM was defined to 1 when sACN was compiled (the default on non-embedded
+     * platforms), and the configuration you pass to sacn_universe_discovery_create() has source_count_max and
+     * universes_per_source_max set to #SACN_UNIVERSE_DISCOVERY_INFINITE, this callback will never be called and may be
+     * set to NULL.
+     *
+     * if #SACN_DYNAMIC_MEM was defined to 0 when sACN was compiled, source_count_max is ignored and
```
ChristianReese, line 97: If

ChristianReese: source_count_max and universes_per_source_max are ignored

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,327 @@
...
+    /********* Optional values **********/
+
+    /** The maximum number of sources this listener will record.  It is recommended that applications using dynamic
+       memory use #SACN_UNIVERSE_DISCOVERY_INFINITE for this value. This parameter is ignored when configured to use
+       static memory -- #SACN_UNIVERSE_DISCOVERY_MAX_SOURCES is used instead.*/
+    size_t source_count_max{SACN_UNIVERSE_DISCOVERY_INFINITE};
+
+    /** The maximum number of universes this listener will record for a source.  It is recommended that applications
+       using dynamic memory use #SACN_UNIVERSE_DISCOVERY_INFINITE for this value. This parameter is ignored when
+       configured to use static memory -- #SACN_UNIVERSE_DISCOVERY_MAX_SOURCES is used instead.*/
```
ChristianReese, line 124: #SACN_UNIVERSE_DISCOVERY_MAX_UNIVERSES_PER_SOURCE

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,327 @@
...
+ * @endcond
+ */
+
+/**
+ * @brief Start a new sACN Universe Discovery listener.
+ *
+ * Note that a listener is considered as successfully created if it is able to successfully use any of the
+ * network interfaces passed in.  This will only return #kEtcPalErrNoNetints if none of the interfaces work.
+ *
+ * @param[in] Settings Configuration parameters for the sACN Universe Discovery listener to be created.
```
ChristianReese, line 204: settings

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,327 @@
...
+ * @param[in, out] netints Optional. If non-NULL, this is the list of interfaces the application wants to use, and the
```
ChristianReese, line 205: @param[in] notify_handler

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,327 @@
...
+
+/**
+ * @brief Start a new sACN Universe Discovery listener with default settings.
+ *
+ * This is an override of Startup that doesn't require a Settings parameter, since the fields in that
+ * structure are completely optional.
+ *
+ * Note that a listener is considered as successfully created if it is able to successfully use any of the
+ * network interfaces passed in.  This will only return #kEtcPalErrNoNetints if none of the interfaces work.
+ *
```
ChristianReese, line 234: @param[in] notify_handler

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,327 @@
...
+ * After this call completes successfully, the listener will continue as if nothing had changed. New sources could be
+ * discovered, or old sources could expire.
+ * If this call fails, the caller must call Shutdown() for this class instance, because it may
+ * be in an invalid state.
+ *
+ * Note that the networking reset is considered successful if it is able to successfully use any of the
+ * network interfaces passed in.  This will only return #kEtcPalErrNoNetints if none of the interfaces work.
+ *
+ * @param[in, out] netints Optional. If non-NULL, this is the list of interfaces the application wants to use, and the
```
ChristianReese, line 278: Update this, since netints is a vector.

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,327 @@
...
+ * operation_succeeded flags are filled in.  If NULL, all available interfaces are tried.
+ * @param[in, out] num_netints Optional. The size of netints, or 0 if netints is NULL.
```
ChristianReese, line 280: Remove

ChristianReese (approved):
Just a few minor adjustments

include/sacn/universe_discovery.h
```diff
@@ -0,0 +1,253 @@
...
+  {                                                                                               \
+    {NULL, NULL, NULL, NULL}, SACN_UNIVERSE_DISCOVERY_INFINITE, SACN_UNIVERSE_DISCOVERY_INFINITE, \
+  }
+
+void sacn_universe_discovery_config_init(SacnUniverseDiscoveryConfig* config);
+
+etcpal_error_t sacn_universe_discovery_create(const SacnUniverseDiscoveryConfig* config,
+                                              sacn_universe_discovery_t* handle, SacnMcastInterface* netints,
+                                              size_t num_netints);
+etcpal_error_t sacn_universe_discovery_destroy(sacn_universe_discovery_t handle);
```
ChristianReese, line 240: I've seen other destroy functions return void. Can we do the same here?

nwagner: Must have been a copy paste error.  I agree.

nwagner:
A code review will always improve your code.  It always amazes me how many little things get missed with one set of eyes.

ChristianReese (approved):

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,399 @@
...
+ * std::vector<SacnMcastInterface> my_netints;
+ * // Assuming my_netints is initialized by the application...
+ *
+ * UniverseDiscovery my_universe_discovery;
+ * MyNotifyHandler my_notify;  // MyNotifyHandler is an application-defined class for implementing notifications.
+ *
+ * // If you want to specify specific network interfaces to use:
+ * etcpal::Error startup_result = my_universe_discovery.Startup(my_notify, my_netints);
+ * // Or, if you just want to use all network interfaces:
+ * etcpal::Error startup_result = my_universe_discovery.Startup(my_notify, std::vector<SacnMcastInterface>());
```
samkearney, line 67: Can we have the netints vector be a default arg to an empty vector?

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,399 @@
...
+ * my_universe_discovery.Shutdown();
+ * 
+ * // During application shutdown, everything can be cleaned up by calling sacn::Deinit.
+ * sacn::Deinit();
+ * @endcode
+ *
+ * Callback demonstrations:
+ * @code
+ * void MyNotifyHandler::HandleUpdateSource(const etcpal::Uuid& cid, const std::string& name,
+ *                                          const uint16_t* sourced_universes, size_t num_sourced_sources)
```
samkearney, line 86: num_sourced_sources should be num_sourced_universes

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,399 @@
...
+ * }
+ *
+ * void MyNotifyHandler::HandleMemoryLimitExceeded()
+ * {
+ *   std::cout << "Universe discovery: Source/universe limit exceeded!\n";
+ * }
+ * @endcode
+ */
+
+class UniverseDiscovery
```
samkearney, line 114: Class names should generally be nouns. How about UniverseListener?

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,399 @@
...
+     * The list of sourced universes is guaranteed by the protocol to be numerically sorted.
+     *
+     * @param[in] cid The CID of the source.
+     * @param[in] name The UTF-8 name string.
+     * @param[in] sourced_universes Numerically sorted array of the currently sourced universes.  Will be NULL if the
+     * source is not currently transmitting any universes.
+     * @param[in] num_sourced_universe Size of the sourced_universes array.  Will be 0 if the source is not currently
+     * transmitting any universes.
+     */
+    virtual void HandleUpdateSource(const etcpal::Uuid& cid, const std::string& name, const uint16_t* sourced_universes,
```
samkearney, line 146: For the C++ API, can we translate sourced_universes into a vector rather than pointer and length for nicer syntax?

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,399 @@
...
+     *
+     * If #SACN_DYNAMIC_MEM was defined to 0 when sACN was compiled, source_count_max and universes_per_source_max are
+     * ignored and #SACN_UNIVERSE_DISCOVERY_MAX_SOURCES and #SACN_UNIVERSE_DISCOVERY_MAX_UNIVERSES_PER_SOURCE are used
+     * instead.
+     *
+     * This callback is rate-limited: it will only be called when the first universe discovery packet is received that
+     * takes the module beyond a memory limit.  After that, it will not be called until the number of sources or
+     * universes has dropped below the limit and hits it again.
+     */
+    virtual void HandleMemoryLimitExceeded() = 0;
```
samkearney, line 173: This should be optional (not pure virtual). The comment about setting it to NULL is probably copied and pasted from a C callback and should be updated.

include/sacn/cpp/universe_discovery.h
samkearney, line 146: How about HandleSourceUpdated() to remain consistent with the semantic theme of "Handle (an event that occurred)"?

include/sacn/cpp/universe_discovery.h
```diff
@@ -0,0 +1,399 @@
...
+  struct Settings
+  {
+    /********* Required values **********/
+
+    /********* Optional values **********/
+
+    /** The maximum number of sources this listener will record.  It is recommended that applications using dynamic
+       memory use #SACN_UNIVERSE_DISCOVERY_INFINITE for this value. This parameter is ignored when configured to use
+       static memory -- #SACN_UNIVERSE_DISCOVERY_MAX_SOURCES is used instead.*/
+    size_t source_count_max{SACN_UNIVERSE_DISCOVERY_INFINITE};
```
samkearney, line 189: Let's make these ints rather than size_t

include/sacn/universe_discovery.h
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * @{
+ */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** A handle to an sACN universe discovery listener. */
+typedef int sacn_universe_discovery_t;
```
samkearney, line 136: How about sacn_universe_listener_t?

include/sacn/universe_discovery.h
```diff
@@ -0,0 +1,252 @@
...
+/** An invalid sACN universe discovery handle value. */
+#define SACN_UNIVERSE_DISCOVERY_INVALID -1
+
+/**
+ * @brief Constant for "infinite" when listening for sources or universes on a source.
+ *
+ * When using dynamic memory, this constant can be passed in when creating a universe discovery listener.
+ * It represents an infinite number of sources or universes on a source.
+ */
+#define SACN_UNIVERSE_DISCOVERY_INFINITE 0
```
samkearney, line 146: When the parameters this is used for are changed to ints, this should probably become -1.

include/sacn/universe_discovery.h
```diff
@@ -0,0 +1,252 @@
...
+ * @param[in] handle Handle to the listening universe discovery instance.
+ * @param[in] cid The CID of the source.
+ * @param[in] name The null-terminated UTF-8 string.
+ * @param[in] sourced_universes Numerically sorted array of the currently sourced universes.  Will be NULL if the source
+ * is not currently transmitting any universes.
+ * @param[in] num_sourced_universes Size of the sourced_universes array.  Will be 0 if the source is not currently
+ * transmitting any universes.
+ * @param[in] context Context pointer that was given at the creation of the universe discovery instance.
+ */
+typedef void (*SacnUniverseDiscoveryUpdateSourceCallback)(sacn_universe_discovery_t handle, const EtcPalUuid* cid,
```
samkearney, line 165: How about SacnUniverseDiscoverySourceUpdatedCallback?

include/sacn/universe_discovery.h
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * @param[in] handle Handle to the listening universe discovery instance.
+ * @param[in] context Context pointer that was given at the creation of the universe discovery instance.
+ */
+typedef void (*SacnUniverseDiscoveryLimitExceededCallback)(sacn_universe_discovery_t handle, void* context);
+
+/** A set of callback functions that the library uses to notify the application about universe discovery events. */
+typedef struct SacnUniverseDiscoveryCallbacks
+{
+  SacnUniverseDiscoveryUpdateSourceCallback source_update;   /**< Required */
```
samkearney, line 203: How about source_updated?

src/sacn/private/opts.h
```diff
@@ -320,6 +320,79 @@
...
+ * @{
+ */
+
+/**
+ * @brief The priority of the sACN universe discovery thread.
+ *
+ * This is usually only meaningful on real-time systems.
+ */
+#ifndef SACN_UNIVERSE_DISCOVERY_THREAD_PRIORITY
+#define SACN_UNIVERSE_DISCOVERY_THREAD_PRIORITY ETCPAL_THREAD_DEFAULT_PRIORITY
```
samkearney, line 343: Are we starting another thread for universe discovery? Would it not work/make sense to piggyback on one of the receiver threads?

samkearney (changes requested):
I'm still on the fence about whether this should be integrated with one of the receiver APIs, but this seems fine. I think we should rename the files and APIs from "Universe Discovery" to "Universe Listener", or to a similar noun that refers to what the instances are that the user creates.

nwagner:
@samkearney , I get what you're saying about changing the noun.  I noticed I was starting to use the term listener in the comments, and that could have been a clue. :)  I'm not fond of UniverseListener -- that might be a little ambiguous with the receiver.  How about Discovery Listener?

include/sacn/cpp/universe_discovery.h
nwagner: I'm on the fence --The main reason I didn't do this was that it would be copying memory, and if the library was set to static memory this would be a conflict.  This is why std::span would be so darn useful.  Given that this api will probably only be used in the dynamic memory case, I could easily be convinced.  Do you still want this?

include/sacn/universe_discovery.h
nwagner: I may change this to ints, but I think it's useful to keep this at 0.  Then we get a default of "infinite" for free my memsetting the config structure to 0.  It doesn't make sense to have a 0 limit anyway.

src/sacn/private/opts.h
nwagner: It may.  I can add a note in here for Christian to remove if he decides to do that in implementation.  Good catch.

samkearney:
Discovery Listener is a bit too redundant/generic. Here are some other options:
* Universe Detector
* Source Detector
* Universe Finder
* Source Finder
* Source Listener

include/sacn/cpp/universe_discovery.h
samkearney: RDMnet uses allocating containers in the C++ API regardless of the library's memory configuration. I was just looking to copy that pattern. It's up to you.

nwagner: That's reasonable -- I've implemented that change.

include/sacn/cpp/universe_discovery.h
nwagner: SourceDetector is what I landed on.  Thanks for the naming idea!

include/sacn/universe_discovery.h
nwagner: sacn_source_detector_t

nwagner:
Ok, this should resolve the last of the comments, including the change from Universe Discovery to Source Detector.  Let me know if you need any more changes, or if I can merge.

ChristianReese (approved):

include/sacn/cpp/source_detector.h
```diff
@@ -0,0 +1,402 @@
...
+  void Shutdown();
+  etcpal::Error ResetNetworking(std::vector<SacnMcastInterface>& netints = kAllInterfaces);
+
+  constexpr Handle handle() const;
+
+private:
+  SacnSourceDetectorConfig TranslateConfig(const Settings& settings, NotifyHandler& notify_handler);
+
+  /** An internal shortcut for selecting all network interfaces. **/
+  static std::vector<SacnMcastInterface> kAllInterfaces;
```
samkearney, line 218: Does this really compile without being defined somewhere?

src/sacn/common.c
```diff
@@ -76,6 +77,7 @@ etcpal_error_t sacn_init(const EtcPalLogParams* log_params)
     bool source_initted = false;
     bool merger_initted = false;
     bool merge_receiver_initted = false;
+    bool universe_discovery_initted = false;
```
samkearney, line 80: source_detector_initted

tests/unit/api/cpp/source_detector/test_cpp_source_detector.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+  void TearDown() override
+  {
+    sacn_source_detector_deinit();
+    sacn_mem_deinit();
+  }
+};
+
+TEST_F(TestUniverseDiscovery, TestFoo)
+{
```
samkearney, line 62: Change UniverseDiscovery to SourceDetector in these test fixtures. I'd also recommend just instantiating a default-constructed sacn::SourceDetector in this test fixture placeholder to make sure it compiles.

include/sacn/cpp/source_detector.h
nwagner: You're right -- that's bizarre.  I was focusing on how to make the default parameter work, but this doesn't make sense.  I'm wondering if I should remove this entirely and make the caller pass in the empty vector again.  The difficulty is that this is an in/out parameter, so you can't just have `foo(std::vector<SacnMcastInterface>& param = std::vector<SacnMcastInterface>());`  What do you think, @samkearney?

samkearney: > The difficulty is that this is an in/out parameter, so you can't just have `foo(std::vector<SacnMcastInterface>& param = std::vector<SacnMcastInterface>());`

Wait, why not?

nwagner: Because it won't compile. :)

nwagner: The exact error is:
`3>D:\Projects\sACN\include\sacn/cpp/source_detector.h(206,66): error C2440: 'default argument': cannot convert from 'std::vector<SacnMcastInterface,std::allocator<SacnMcastInterface>>' to 'std::vector<SacnMcastInterface,std::allocator<SacnMcastInterface>> &'
3>D:\Projects\sACN\include\sacn/cpp/source_detector.h(206,66): message : A non-const reference may only be bound to an lvalue
`

samkearney: Oh, I see. I understand why that language limitation exists, although it hurts us in this case.

OK, how about the kAllInterfaces member, but make it non-static?

nwagner: `2>D:\Projects\sACN\include\sacn/cpp/source_detector.h(206,66): error C2648: 'sacn::SourceDetector::kAllInterfaces': use of member as default parameter requires static member
2>D:\Projects\sACN\include\sacn/cpp/source_detector.h(218): message : see declaration of 'sacn::SourceDetector::kAllInterfaces'
2>D:\Projects\sACN\include\sacn/cpp/source_detector.h(206,66): error C2355: 'this': can only be referenced inside non-static member functions or non-static data member initializers
`
Keep the ideas coming!  I thought I tried everything, but you might see something I missed.

samkearney: More overloads of Startup():
```cpp
etcpal::Error Startup(NotifyHandler& notify_handler);
etcpal::Error Startup(const Settings& settings, NotifyHandler& notify_handler);
etcpal::Error Startup(NotifyHandler& notify_handler, std::vector<SacnMcastInterface>& netints);
etcpal::Error Startup(const Settings& settings, NotifyHandler& notify_handler, std::vector<SacnMcastInterface>& netints);
```

nwagner: I was trying to find something more elegant, but I guess this is probably the best bet under the circumstances.  I'll make this change.  Thanks!

nwagner: Ok, my updates should be done.  Let me know if you have anything else.

samkearney (approved):

PR #13, opened 2020-11-17
Giving thanks for receiver

ChristianReese:
Additional November sprint work that has been done. More work to do this week. PRing early so that we can review this sooner.


ChristianReese:
At this point I believe I have covered all requested changes as of 11/18, on the public-facing side. There may be one more pending tomorrow's conversation - otherwise, this should be everything needed for the API review. Implementation needs a cleanup, but that doesn't matter for the API review.

ChristianReese:
I believe everything is in there now. Should be good to review.

include/sacn/receiver.h
```diff
@@ -204,6 +155,24 @@ typedef void (*SacnUniverseDataCallback)(sacn_receiver_t handle, uint16_t univer
...
+ * @brief Notify that a receiver's sampling period has ended.
+ *
+ * @param[in] handle Handle to the receiver instance for which the sampling period ended.
+ * @param[in] universe The universe this receiver is monitoring.
+ * @param[in] context Context pointer that was given at the creation of the receiver instance.
+ */
+typedef void (*SacnSamplingPeriodEndedCallback)(sacn_receiver_t handle, uint16_t universe, void* context);
+
+/**
+ * @brief Notify that a receiver's sampling period has begun.
```
nwagner, line 168: Nitpick again, start before end

include/sacn/receiver.h
```diff
@@ -246,11 +215,12 @@ typedef void (*SacnSourceLimitExceededCallback)(sacn_receiver_t handle, uint16_t
...
 {
-  SacnSourcesFoundCallback sources_found;                /**< Required */
-  SacnUniverseDataCallback universe_data;                /**< Required */
-  SacnSourcesLostCallback sources_lost;                  /**< Required */
-  SacnSourcePapLostCallback source_pap_lost;             /**< Optional */
-  SacnSourceLimitExceededCallback source_limit_exceeded; /**< Optional */
+  SacnUniverseDataCallback universe_data;                    /**< Required */
+  SacnSourcesLostCallback sources_lost;                      /**< Required */
+  SacnSamplingPeriodEndedCallback sampling_period_ended;     /**< Required */
+  SacnSamplingPeriodStartedCallback sampling_period_started; /**< Optional */
```
nwagner, line 221: Nitpick

nwagner (approved):
A couple of minor tweaks, but this is looking great!

nwagner (approved):

include/sacn/receiver.h
```diff
@@ -246,11 +215,12 @@ typedef void (*SacnSourceLimitExceededCallback)(sacn_receiver_t handle, uint16_t
...
 {
-  SacnSourcesFoundCallback sources_found;                /**< Required */
-  SacnUniverseDataCallback universe_data;                /**< Required */
-  SacnSourcesLostCallback sources_lost;                  /**< Required */
-  SacnSourcePapLostCallback source_pap_lost;             /**< Optional */
-  SacnSourceLimitExceededCallback source_limit_exceeded; /**< Optional */
+  SacnUniverseDataCallback universe_data;                    /**< Required */
+  SacnSourcesLostCallback sources_lost;                      /**< Required */
+  SacnSamplingPeriodStartedCallback sampling_period_started; /**< Optional */
+  SacnSamplingPeriodEndedCallback sampling_period_ended;     /**< Required */
```
samkearney, line 221: sampling_period_ended is optional.

src/sacn/private/opts.h
```diff
@@ -253,10 +258,10 @@
 /** @endcond */
 
 /**
- * @brief Currently unused; will be used in the future.
+ * @brief Currently unused; will be used in the future. Keep this set to 1.
```
samkearney, line 261: Instead of telling the user to keep it set to 1, we should just do it:
```
#undef SACN_RECEIVER_MAX_THREADS
#define SACN_RECEIVER_MAX_THREADS 1
```

nwagner (approved):
Barring Sam's requests, this looks good.

include/sacn/cpp/source.h
```diff
@@ -121,6 +121,10 @@ class Source
     /** If true, this sACN source will only send unicast traffic on this universe. Defaults to false. */
     bool send_unicast_only{false};
 
+    /** The initial set of unicast destinations for this universe. This can be changed further by using
+        Source::AddUnicastDestination() and Source::RemoveUnicastDestination(). */
+    const std::vector<EtcPalIpAddr> unicast_destinations;
```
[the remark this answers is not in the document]
ChristianReese: Had to do a little bit of legwork since I'm storing a private vector of `EtcPalIpAddr` since that memory needs to live somewhere. See if you like what I wrote.

include/sacn/cpp/source.h
```diff
@@ -133,22 +134,26 @@ class Source
     UniverseSettings(uint16_t universe_id, const uint8_t* new_values_buffer, size_t new_values_size);
 
     bool IsValid() const;
-  };
 
+  private:
+    std::vector<EtcPalIpAddr> c_unicast_destinations;
```
nwagner, line 139: I'm not understanding why this is needed. Internally, the settings would be translated & used, right?  Not re-stored here..

include/sacn/cpp/source.h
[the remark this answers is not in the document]
samkearney: @ChristianReese  Take a look at how it's done in RDMnet: https://github.com/ETCLabs/RDMnet/blob/develop/include/rdmnet/cpp/device.h#L479

include/sacn/cpp/source.h
```diff
@@ -133,22 +134,26 @@ class Source
...
+  Source& operator=(Source&& other) = default; /**< Move a source instance. */
 
   etcpal::Error Startup(const Settings& settings);
   etcpal::Error Startup(const Settings& settings, std::vector<SacnMcastInterface>& netints);
   void Shutdown();
 
   etcpal::Error ChangeName(const std::string& new_name);
 
-  etcpal::Error AddUniverse(const UniverseSettings& settings);
+  etcpal::Error AddUniverse(UniverseSettings& settings);
```
nwagner, line 156: I don't think you need to remove the const.

include/sacn/cpp/source.h
```diff
@@ -174,7 +179,7 @@ class Source
 
 private:
   SacnSourceConfig TranslateConfig(const Settings& settings);
-  SacnSourceUniverseConfig TranslateUniverseConfig(const UniverseSettings& settings);
+  SacnSourceUniverseConfig TranslateUniverseConfig(UniverseSettings& settings);
```
nwagner, line 182: Don't remove the const

include/sacn/cpp/source.h
```diff
@@ -314,7 +320,7 @@ inline etcpal::Error Source::ChangeName(const std::string& new_name)
  * If the source is not marked as unicast_only, the source will add the universe to its sACN Universe
  * Discovery packets.
 
- * @param[in] config Configuration parameters for the universe to be added.
+ * @param[in,out] config Configuration parameters for the universe to be added. Private translation data is updated.
```
nwagner, line 323: I don't think this needs to be an out

include/sacn/cpp/source.h
```diff
@@ -323,7 +329,7 @@ inline etcpal::Error Source::ChangeName(const std::string& new_name)
  * @return #kEtcPalErrNoMem: No room to allocate additional universe.
  * @return #kEtcPalErrSys: An internal library or system call error occurred.
  */
-inline etcpal::Error Source::AddUniverse(const UniverseSettings& settings)
+inline etcpal::Error Source::AddUniverse(UniverseSettings& settings)
```
nwagner, line 332: ditto

nwagner (changes requested):
I'm not a fan of the c_unicast_destinations changes.  Let's talk when I'm back after 12.

include/sacn/cpp/source.h
ChristianReese: I thought that the source API would need to keep the data, but then I realized it will just copy the unicast dests into its internal state that is shared with the add/remove dest functions. Good catch, thanks!

include/sacn/cpp/source.h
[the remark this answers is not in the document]
ChristianReese: Good to know for the future. Nick made me realize that this memory will only be used temporarily, but I've bookmarked this for future reference.

include/sacn/cpp/source.h
ChristianReese: Wait... the memory still needs somewhere to live just to get passed to the function. Let's talk.

samkearney: @ChristianReese like I said, take a look at how it's done in RDMnet: https://github.com/ETCLabs/RDMnet/blob/develop/include/rdmnet/cpp/device.h#L479

ChristianReese: Going with that.

include/sacn/cpp/source.h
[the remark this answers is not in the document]
ChristianReese: Backtracking again. The memory still needs a place to live so I'm going with the RDMnet implementation.

ChristianReese:
Went with the RDMnet approach. See what you guys think.

nwagner (approved):
This looks much better -- Thanks!

samkearney (approved):

ChristianReese:
Ignoring the build issues for now.

PR #17, opened 2020-12-01
Add documentation of per-channel priority

docsteer:
Add a Markdown version of the ETC documentation of per-channel priority


samkearney (commented):
Can we change the following terminology to align better with what we use in the new library docs:

* Sink -> Receiver
* Data loss behavior -> Source loss behavior

Otherwise, looks good. Thanks!

docs/pages/per_address_priority.md
```diff
@@ -0,0 +1,54 @@
...
+16. When a receiver detects that a source has given up control of an address (by setting the priority for that address to 0), the receiver will give control to the source with next highest priority. If no source is available, source loss behavior for that address shall apply.
+
+# Summary
+In summary, priority may be specified on a per packet basis in sACN. This ETC extension allows priorities to be provided on a per-address basis using the alternate start code of 0xDD.
+
+With this extension, taking control is accomplished by setting a non-zero priority for an address.  Giving up control is the reverse, setting the priority to 0.
+
+Overall, receivers wait for a short period on start up to see if the alternate start code priority information is being provided before they start accepting data.
+
+If no alternate start code priority information arrives, they use the packet priority.  If the packet is a Draft sACN packet and it specifies that it does not carry a priority, by placing a zero in the field.
```
nwagner, line 54: I don't think we need to reference draft any more.

nwagner (approved):
Very nice!

docs/pages/per_address_priority.md
samkearney: Agreed

PR #27, opened 2021-01-04
sACN Receiver: Implement ip_supported Setting

ChristianReese:
This is an implementation of the ip_supported setting in the receiver configuration. This has also been added to the unit tests for change universe.

### DIFF
--- a/src/sacn/receiver.c
+++ b/src/sacn/receiver.c
@@ -438,18 +438,28 @@ etcpal_error_t sacn_receiver_change_universe(sacn_receiver_t handle, uint16_t ne
     // Update the receiver's socket and subscription.
     if (res == kEtcPalErrOk)
     {
-      sacn_remove_receiver_socket(receiver->thread_id, receiver->ipv4_socket, false);
-      sacn_remove_receiver_socket(receiver->thread_id, receiver->ipv6_socket, false);
-      res = sacn_add_receiver_socket(receiver->thread_id, kEtcPalIpTypeV4, new_universe_id, receiver->netints,
-                                     receiver->num_netints, &receiver->ipv4_socket);
-      if (res == kEtcPalErrNoNetints)
-        res = kEtcPalErrOk;  // Try IPv6.
+      if (receiver->ipv4_socket != ETCPAL_SOCKET_INVALID)
+        sacn_remove_receiver_socket(receiver->thread_id, &receiver->ipv4_socket, false);
+      if (receiver->ipv6_socket != ETCPAL_SOCKET_INVALID)
+        sacn_remove_receiver_socket(receiver->thread_id, &receiver->ipv6_socket, false);
+
+      if ((receiver->ip_supported == kSacnIpV4Only) || (receiver->ip_supported == kSacnIpV4AndIpV6))
+      {
+        res = sacn_add_receiver_socket(receiver->thread_id, kEtcPalIpTypeV4, new_universe_id, receiver->netints,
+                                       receiver->num_netints, &receiver->ipv4_socket);
+
+        if ((res == kEtcPalErrNoNetints) && (receiver->ip_supported == kSacnIpV4AndIpV6))
+          res = kEtcPalErrOk;  // Try IPv6.
+      }
     }
 
     if (res == kEtcPalErrOk)
     {
-      res = sacn_add_receiver_socket(receiver->thread_id, kEtcPalIpTypeV6, new_universe_id, receiver->netints,
-                                     receiver->num_netints, &receiver->ipv6_socket);
+      if ((receiver->ip_supported == kSacnIpV6Only) || (receiver->ip_supported == kSacnIpV4AndIpV6))
+      {
+        res = sacn_add_receiver_socket(receiver->thread_id, kEtcPalIpTypeV6, new_universe_id, receiver->netints,
+                                       receiver->num_netints, &receiver->ipv6_socket);
+      }
     }
 
     // Update receiver key and position in receiver_state.receivers_by_universe.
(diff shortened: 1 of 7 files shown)


tests/unit/api/c/receiver/test_receiver.cpp
```diff
@@ -66,6 +67,109 @@ class TestReceiver : public ::testing::Test
...
+      case kSacnIpV6Only:
+        EXPECT_EQ(sacn_remove_receiver_socket_fake.call_count, 1u);
+        EXPECT_EQ(sacn_add_receiver_socket_fake.call_count, 2u);
+        break;
+      case kSacnIpV4AndIpV6:
+        EXPECT_EQ(sacn_remove_receiver_socket_fake.call_count, 2u);
+        EXPECT_EQ(sacn_add_receiver_socket_fake.call_count, 4u);
+    }
+
+    clear_term_set_list_fake.custom_fake = nullptr;
```
samkearney, line 167: If these fakes are not going to be used again before the test is over, this isn't necessary - sacn_source_loss_reset_all_fakes() in the SetUp() method will take care of it.

tests/unit/api/c/receiver/test_receiver.cpp
```diff
@@ -66,6 +67,109 @@ class TestReceiver : public ::testing::Test
     sacn_receiver_deinit();
     sacn_mem_deinit();
   }
+
+  void TestChangeUniverse(sacn_ip_support_t ip_supported)
```
samkearney, line 71: This is kind of a long function - one idea that might make this test code easier to understand, is if you put all the common setup code at the beginning of the function in its own "Setup" function, e.g. SetupUniverseChangeTest(). Then transfer the code in the big switch statement below to each individual TEST_F() function. Just a thought.

samkearney (approved):

src/sacn/receiver.c
```diff
@@ -774,24 +786,35 @@ etcpal_error_t assign_receiver_to_thread(SacnReceiver* receiver, const SacnRecei
 
   SACN_ASSERT(assigned_thread);
 
-  etcpal_error_t res = sacn_add_receiver_socket(receiver->thread_id, kEtcPalIpTypeV4, config->universe_id,
-                                                receiver->netints, receiver->num_netints, &receiver->ipv4_socket);
-  if (res == kEtcPalErrNoNetints)
-    res = kEtcPalErrOk;  // Try IPv6.
+  etcpal_error_t res = kEtcPalErrOk;
+
+  if ((receiver->ip_supported == kSacnIpV4Only) || (receiver->ip_supported == kSacnIpV4AndIpV6))
```
nwagner, line 791: This if clause and the next could be turned into a single utility function.  It's used here and above.

src/sacn/receiver.c
```diff
@@ -774,24 +786,35 @@ etcpal_error_t assign_receiver_to_thread(SacnReceiver* receiver, const SacnRecei
...
   }
 
   if (res == kEtcPalErrOk && !assigned_thread->running)
   {
     res = start_receiver_thread(assigned_thread);
     if (res != kEtcPalErrOk)
     {
-      sacn_remove_receiver_socket(receiver->thread_id, receiver->ipv4_socket, true);
-      sacn_remove_receiver_socket(receiver->thread_id, receiver->ipv6_socket, true);
+      if (receiver->ipv4_socket != ETCPAL_SOCKET_INVALID)
```
nwagner, line 814: Likewise, you are doing this same cleanup in a spots, it could be a utility function.

src/sacn/sockets.c
```diff
@@ -321,14 +321,17 @@ etcpal_error_t sacn_add_receiver_socket(sacn_thread_id_t thread_id, etcpal_iptyp
...
 {
-  SACN_ASSERT(socket != ETCPAL_SOCKET_INVALID);
+  SACN_ASSERT(socket != NULL);
+  SACN_ASSERT(*socket != ETCPAL_SOCKET_INVALID);
 
   SacnRecvThreadContext* context = get_recv_thread_context(thread_id);
   SACN_ASSERT(context);
 
-  cleanup_socket(context, socket, close_now);
+  cleanup_socket(context, *socket, close_now);
```
nwagner, line 332: Normally, I don't try to rely on asserts to guarantee validity.  I would typically check the validity of context & socket at runtime.  Only with good analysis would I consider doing this in a function that has critical performance needs.

tests/unit/api/c/receiver/test_receiver.cpp
```diff
@@ -66,6 +67,109 @@ class TestReceiver : public ::testing::Test
...
+                                                       etcpal_socket_t* socket) {
+          EXPECT_NE(ip_type, kEtcPalIpTypeInvalid);
+          EXPECT_EQ(universe, CHANGE_UNIVERSE_WORKS_SECOND_UNIVERSE);
+          EXPECT_NE(get_recv_thread_context(thread_id), nullptr);
+          EXPECT_NE(socket, nullptr);
+          return kEtcPalErrOk;
+        };
+    }
+
+    etcpal_error_t change_universe_result =
```
nwagner, line 147: If you do follow Sam's suggestion, lines 147 on could be refactored into a function that your individual test functions can call (e.g. PerformUniverseChangeTest)

nwagner (approved):

src/sacn/sockets.c
samkearney: Assert should be used when a function has preconditions that would only be violated if there was a bug in the library. I believe this qualifies. We should fail early and hard in the presence of bugs and not use runtime checks to cover them up.

ChristianReese: When I wrote this part, I was also faced with the decision of using assert or runtime checks, and I went asserts for the reasons Sam describes. I felt that this way would provide stricter error checking and I couldn't think of any drawbacks. Are you okay with keeping this as is @nwagner ?

samkearney (approved):

nwagner (approved):

src/sacn/sockets.c
ChristianReese: We talked offline and decided to leave this as-is.

PR #28, opened 2021-01-05
Provide more specific error statuses in SacnMcastInterface::status

ChristianReese:
I ended up adjusting a few things here. SACN_MAX_NETINTS now refers to the maximum number of total interfaces that can be processed by sACN, instead of just the valid ones (although it was already partially the former). Now sACN is tracking all interfaces because it is tracking the error codes of the failed interfaces in addition to the valid ones.

SACN_MAX_NETINTS has the same documentation but let me know if you want it adjusted: "The maximum number of network interfaces that can used by the sACN library. Meaningful only if #SACN_DYNAMIC_MEM is defined to 0."

I am passing along the error statuses directly from the failed netint test operations. I'm currently not doing any filtering of these statuses beyond that, but let me know if there should be.

### DIFF
--- a/src/sacn/sockets.c
+++ b/src/sacn/sockets.c
@@ -513,11 +513,16 @@ etcpal_error_t sacn_validate_netint_config(SacnMcastInterface* netints, size_t n
 
     for (SacnMcastInterface* netint = netints; netint < (netints + num_netints); ++netint)
     {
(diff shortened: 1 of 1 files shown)


ChristianReese:
I'm going to fix the automated build failures after lunch.

samkearney (approved):

nwagner (approved):

PR #29, opened 2021-01-06
DMX Merger: Implemented New Update Functions

ChristianReese:
The new update functions, for updating levels, PAPs, and universe priority, have been implemented. The merger unit tests have also been updated. This implementation also accounts for the rule that a source must have at least a level and either a PAP or universe priority at a particular slot for it to be factored into the merge at that slot.


ChristianReese:
Made some last minute adjustments. Should be ready for review now.

nwagner (approved):

src/sacn/dmx_merger.c
```diff
@@ -974,6 +966,44 @@ MergerState* construct_merger_state(sacn_dmx_merger_t handle, const SacnDmxMerge
...
+  // Look up the source state.
+  if ((result == kEtcPalErrOk) && source_state)
+  {
+    my_source_state = etcpal_rbtree_find(&my_merger_state->source_state_lookup, &source);
+
+    if (!my_source_state)
+      result = kEtcPalErrNotFound;
+  }
+
+  if (merger_state)
```
samkearney, line 999: Since the function returns an error code if one of the two states is not found, I recommend not writing to either of the output pointers if one of the lookups fail:
```
if (result == kEtcPalErrOk)
{
  if (merger_state)
    *merger_state = my_merger_state;
  if (source_state)
    *source_state = my_source_state;
}
```

tests/unit/api/c/dmx_merger/test_dmx_merger.cpp
```diff
@@ -75,15 +76,15 @@ class TestDmxMerger : public ::testing::Test
...
+    kUpdate,        // Merge using the sacn_dmx_merger_update functions.
+    kStopSourcePap  // Merge using the sacn_dmx_merger_update functions, then call
+                    // sacn_dmx_merger_remove_paps on the second source.
   };
 
-  void TestMerge(uint8_t priority_1, const uint8_t* values_1, uint16_t values_1_count,
-                 const uint8_t* address_priorities_1, uint16_t address_priorities_1_count, uint8_t priority_2,
-                 const uint8_t* values_2, uint16_t values_2_count, const uint8_t* address_priorities_2,
-                 uint16_t address_priorities_2_count, MergeTestType merge_type)
+  void TestMerge(const std::optional<int>& priority_1, const uint8_t* levels_1, uint16_t levels_1_count,
```
samkearney, line 84: This function has a ton of parameters, and feels like it's doing too many things. Does this really need to be a generic function / is the MergeTestType enum necessary? I've found that unit tests are one place where some repetition can be OK and can make the test code easier to understand. In a previous review we replaced this pattern with refactoring common setup code into its own setup function, then having the unique test code in each individual TEST_F() function.

nwagner (approved):

tests/unit/api/c/dmx_merger/test_dmx_merger.cpp
ChristianReese: Let me know what you guys think about the direction I went with this: https://github.com/ETCLabs/sACN/pull/29/commits/b5ab10a11905fff9052a4a6bfe34fa44d284b34b

tests/unit/api/c/dmx_merger/test_dmx_merger.cpp
```diff
@@ -608,139 +599,469 @@ TEST_F(TestDmxMerger, GetSourceWorks)
   EXPECT_NE(sacn_dmx_merger_get_source(merger_handle_, source_handle_2), nullptr);
 }
 
-TEST_F(TestDmxMerger, UpdateSourceDataMergesLevels)
+TEST_F(TestDmxMergerUpdate, MergesLevels)
```
nwagner, line 602: On a few of of these tests, you could create variants that do the update operations in a different order to make sure you get the same result.  For example, you could do source 2 first, or interleave setting their priorities and levels.

nwagner (approved):

tests/unit/api/c/dmx_merger/test_dmx_merger.cpp
```diff
@@ -210,9 +101,104 @@ class TestDmxMerger : public ::testing::Test
...
-  uint8_t test_values_ascending_[DMX_ADDRESS_COUNT];
-  uint8_t test_values_descending_[DMX_ADDRESS_COUNT];
+    // Initialize expected merge results.
+    ClearExpectedMergeResults();
+
+    // Initialize test values to be used for levels and PAPs.
+    test_values_ascending_.resize(DMX_ADDRESS_COUNT);
+    test_values_descending_.resize(DMX_ADDRESS_COUNT);
+
+    for (int i = 0; i < DMX_ADDRESS_COUNT; ++i)
```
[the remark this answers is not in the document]
ChristianReese: The problem is that then I wouldn't be able to initialize them with the for loops (the Intellisense complains when I put the loops in the constructor, requiring a modifiable l-value). So then I would need {1, 2, 3, 4, 5, ...} for up to 512 values. Unless if that is still preferable or there's a different way to do this?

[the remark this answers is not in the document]
jtveite: There's a technique I've seen a few places that essentially just does the initialization inside a lambda so that you can do things like for loops while keeping the created variable constant. That could be handy here.
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-lambda-init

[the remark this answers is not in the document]
ChristianReese: Very cool @jtveite ! I didn't know lambdas could be used that way. It looks like this works. Thanks!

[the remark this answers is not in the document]
ChristianReese: Implemented in https://github.com/ETCLabs/sACN/pull/29/commits/93551f5ed28d503c1f06b4d067a021256f788372

tests/unit/api/c/dmx_merger/test_dmx_merger.cpp
```diff
@@ -197,8 +187,19 @@ class TestDmxMergerUpdate : public TestDmxMerger
   int expected_merge_priorities_[DMX_ADDRESS_COUNT];
   sacn_source_id_t expected_merge_winners_[DMX_ADDRESS_COUNT];
 
-  std::vector<uint8_t> test_values_ascending_;
-  std::vector<uint8_t> test_values_descending_;
+  const std::vector<uint8_t> test_values_ascending_ = [&] {
```
nwagner, line 190: This is so cool!

nwagner (approved):

nwagner (approved):

tests/unit/api/c/dmx_merger/test_dmx_merger.cpp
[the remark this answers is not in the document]
samkearney: Cool trick @jtveite , I am gonna start using this in my code

tests/unit/api/c/dmx_merger/test_dmx_merger.cpp
ChristianReese: https://github.com/ETCLabs/sACN/pull/29/commits/b74e25cecfecd7383db78cfd3d2f6de98a49c9f6

nwagner (approved):

PR #32, opened 2021-01-22
sACN Source Implementation (everything besides Reset Networking, unit tests, and example apps)

ChristianReese:
This is where I'm currently at with the Source. This is just the implementation (still need to implement Reset Networking).


src/sacn/source.c
```diff
@@ -595,13 +1232,695 @@ etcpal_error_t sacn_source_reset_networking(sacn_source_t handle, uint16_t unive
...
+}
+
+// Needs lock
+void set_preview_flag(const SacnSource* source, SacnSourceUniverse* universe, bool preview)
+{
+  universe->send_preview = preview;
+  SET_PREVIEW_OPT(universe->null_send_buf, preview);
+  SET_PREVIEW_OPT(universe->pap_send_buf, preview);
+  reset_transmission_suppression(source, universe, true, true);
 }
```
ChristianReese, line 1926: I did do some refactoring to pull code out of this and into other modules. But this file is still fairly big. One idea I have is to make a source_state module, which would be focused on direct operations on the state structs' data. For instance, initialization that takes a config struct. The idea being to include the same functionality as would be in the member functions if each state struct were a C++ class. Let me know your thoughts on that idea. Or if you have other ideas for splitting out this code into other modules, or ideas for new modules.

ChristianReese: Keeping in mind the two benefits of modules - one being smaller, more focused files; two being increased opportunities for unit tests that cover more of the internal functionality.

src/sacn/source.c
```diff
@@ -220,22 +611,76 @@ size_t sacn_source_get_universes(sacn_source_t handle, uint16_t* universes, size
...
+  etcpal_error_t result = kEtcPalErrOk;
+
+  // Verify module initialized.
+  if (!sacn_initialized())
+    result = kEtcPalErrNotInit;
+
+  // Check for invalid arguments.
+  if (result == kEtcPalErrOk)
+  {
+    if (handle == SACN_SOURCE_INVALID)
```
samkearney, line 632: Combine these checks into a single if() block

src/sacn/source.c
```diff
@@ -366,18 +922,61 @@ etcpal_error_t sacn_source_change_synchronization_universe(sacn_source_t handle,
...
```
samkearney, line 938: Combine these checks into a single if() block

src/sacn/source.c
```diff
@@ -595,13 +1232,695 @@ etcpal_error_t sacn_source_reset_networking(sacn_source_t handle, uint16_t unive
...
+}
+
+// Needs lock
+etcpal_error_t start_tick_thread()
+{
+  shutting_down = false;
+  EtcPalThreadParams params = ETCPAL_THREAD_PARAMS_INIT;
+
+  if (etcpal_thread_create(&source_thread_handle, &params, source_thread_function, NULL) != kEtcPalErrOk)
+    return kEtcPalErrSys;
```
samkearney, line 1275: Why not return the error code from etcpal_thread_create()?

src/sacn/source.c
```diff
@@ -595,13 +1232,695 @@ etcpal_error_t sacn_source_reset_networking(sacn_source_t handle, uint16_t unive
...
+    if (sacn_lock())
+    {
+      keep_running_thread = !shutting_down;
+      sacn_unlock();
+    }
+  }
+}
+
+// Takes lock
+int process_internal(bool process_manual)
```
samkearney, line 1323: This function feels like it could be better named, process internal what?

src/sacn/source.c
```diff
@@ -595,13 +1232,695 @@ etcpal_error_t sacn_source_reset_networking(sacn_source_t handle, uint16_t unive
...
+{
+  ETCPAL_UNUSED_ARG(arg);
+
+  bool keep_running_thread = true;
+  int num_thread_based_sources = 0;
+
+  EtcPalTimer interval_timer;
+  etcpal_timer_start(&interval_timer, SOURCE_THREAD_INTERVAL);
+
+  while (keep_running_thread || (num_thread_based_sources > 0))
```
samkearney, line 1307: Hmm, so the thread stops if there are no more thread-based sources, but what starts it again if another thread-based source is added? This logic feels like an overly complex optimization, I think we should start the thread when it is first requested and keep running it until the library is deinitialized.

src/sacn/source.c
```diff
@@ -595,13 +1232,695 @@ etcpal_error_t sacn_source_reset_networking(sacn_source_t handle, uint16_t unive
...
+{
+  ++universe->seq_num;
+  universe->null_send_buf[SACN_SEQ_OFFSET] = universe->seq_num;
+#if SACN_ETC_PRIORITY_EXTENSION
+  universe->pap_send_buf[SACN_SEQ_OFFSET] = universe->seq_num;
+#endif
+}
+
+// Needs lock
+void send_null_data_multicast(const SacnSource* source, SacnSourceUniverse* universe)
```
samkearney, line 1540: Are these functions really necessary if their only job is to call another function?

src/sacn/mem.c
```diff
@@ -660,6 +769,306 @@ void remove_receiver_from_list(SacnRecvThreadContext* recv_thread_context, SacnR
...
+  *dest = etcpal_rbiter_next(unicast_iter);
+  etcpal_rbtree_remove(&universe->unicast_dests, dest_to_remove);
+  FREE_SACN_UNICAST_DESTINATION(dest_to_remove);
+}
+
+// Needs lock
+void remove_sacn_source_universe(SacnSource* source, SacnSourceUniverse** universe, EtcPalRbIter* universe_iter)
+{
+  SacnSourceUniverse* universe_to_remove = *universe;
+  *universe = etcpal_rbiter_next(universe_iter);
```
samkearney, line 951: This won't work. etcpal_rbtree_remove() invalidates all iterators.

src/sacn/source.c
```diff
@@ -595,13 +1232,695 @@ etcpal_error_t sacn_source_reset_networking(sacn_source_t handle, uint16_t unive
...
+    // Unicast destination-specific processing
+    process_unicast_dests(source, universe);
+
+    // Either transmit start codes 0x00 & 0xDD, or terminate and clean up universe
+    if (universe->terminating)
+      process_universe_termination(source, universe, &universe_iter);
+    else
+      process_universe_null_pap_transmission(source, universe);
+
+    if (!universe_incremented)
```
samkearney, line 1409: Good thinking handling the case of universes being removed while iterating, but it looks like the implementation isn't totally fleshed out here - you'll have to assign universe_incremented from the return values of the functions above, and probably change their paramters from SacnSourceUniverse* to SacnSourceUniverse**.

However, as mentioned elsewhere, etcpal_rbtree_remove() invalidates _all_ iterators, so you will have to find another solution for handling removal while iterating in the mem.c functions.

samkearney: rbtrees and iterators might be more trouble than they're worth here because of cases like this, you might consider changing them to arrays.

src/sacn/source.c
samkearney: The only advice I'd give here is that actually writing the unit tests will probably help you figure out what belongs in what module based on how it's easiest to test.

src/sacn/mem.c
```diff
@@ -1082,3 +1490,45 @@ void deinit_source_limit_exceeded_buf(void)
...
 
 #endif  // SACN_DYNAMIC_MEM
+
+etcpal_error_t init_sources(void)
+{
+  etcpal_error_t res = kEtcPalErrOk;
+
+#if SACN_SOURCE_ENABLED
+#if !SACN_DYNAMIC_MEM
+  if (res == kEtcPalErrOk)
```
samkearney, line 1500: Since the mempools don't have a deinit function, I like to take advantage of the fact that kEtcPalErrOk is 0 to shorten the code:
```
res |= etcpal_mempool_init(foo);
res |= etcpal_mempool_init(bar);
res |= etcpal_mempool_init(baz);
// ...
```

src/sacn/private/common.h
```diff
@@ -353,6 +363,84 @@ typedef struct SacnRecvThreadContext
...
+
+  EtcPalRbTree universes;
+  size_t num_active_universes;  // Number of universes to include in universe discovery packets.
+  EtcPalTimer universe_discovery_timer;
+  bool process_manually;
+  sacn_ip_support_t ip_supported;
+  int keep_alive_interval;
+  size_t universe_count_max;
+
+  EtcPalRbTree netints;  // Provides a way to look up netints being used by any universe of this source.
```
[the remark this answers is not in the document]
ChristianReese: When I was implementing universe discovery, I ran into the issue that netints are specific to each universe, but universe discovery encompasses all universes in a source. The approach I decided to go with was to broadcast universe discovery on the set of network interfaces of all universes on the source. To avoid duplicates I ended up going with a rbtree for this. You can see how this is utilized in the send_universe_discovery function.

src/sacn/private/sockets.h
```diff
@@ -41,16 +41,29 @@ typedef struct SacnReadResult
 etcpal_error_t sacn_sockets_init(void);
 void sacn_sockets_deinit(void);
 
-etcpal_error_t sacn_validate_netint_config(SacnMcastInterface* netints, size_t num_netints, size_t *num_valid_netints);
+etcpal_error_t sacn_validate_netint_config(SacnMcastInterface* netints, size_t num_netints, size_t* num_valid_netints);
+#if SACN_DYNAMIC_MEM
+etcpal_error_t sacn_initialize_internal_netints(EtcPalMcastNetintId** internal_netints, size_t* num_internal_netints,
+                                                SacnMcastInterface* app_netints, size_t num_app_netints);
+#else
+etcpal_error_t sacn_initialize_internal_netints(EtcPalMcastNetintId (*internal_netints)[SACN_MAX_NETINTS],
```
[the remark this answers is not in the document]
ChristianReese: I tried that, but unfortunately that causes the build to fail in static memory mode. This was the workaround that I went with in order to allow the use cases to be the same regardless of memory mode. As in, in static memory mode the netints are in an array, but can still be passed in as &netints, just like the pointer in dynamic mode. This originated from me wanting to use a double pointer for netints so I could assign the pointer to calloc, or NULL on failure. Of course this is only utilized in dynamic mode.

src/sacn/source.c
```diff
@@ -83,22 +189,101 @@ void sacn_source_universe_config_init(SacnSourceUniverseConfig* config)
...
+    {
+      source = add_sacn_source(get_next_int_handle(&source_handle_mgr, -1));
+
+      if (!source)
+        result = kEtcPalErrNoMem;
+    }
+
+    if (result == kEtcPalErrOk)
+    {
+      // Initialize the universe discovery send buffer.
```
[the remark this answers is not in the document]
ChristianReese: That was a gray area for me in terms of if that belonged in the mem module or not. But if you think it would fit there, then I can certainly move it over. Or perhaps if I create that source_state module I commented about elsewhere, I could make an init_source function in there?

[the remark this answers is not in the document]
ChristianReese: It does sound nice to have all this in one function instead of two though...

src/sacn/mem.c
```diff
@@ -35,6 +37,72 @@
...
+      free(((SacnSourceUniverse*)ptr)->netints);                                                      \
+    free(ptr);                                                                                        \
+  } while (0)
+#define ALLOC_SACN_SOURCE_NETINT() malloc(sizeof(SacnSourceNetint))
+#define FREE_SACN_SOURCE_NETINT(ptr) free(ptr)
+#define ALLOC_SACN_UNICAST_DESTINATION() malloc(sizeof(SacnUnicastDestination))
+#define FREE_SACN_UNICAST_DESTINATION(ptr) free(ptr)
+#define ALLOC_SACN_SOURCE_RB_NODE() malloc(sizeof(EtcPalRbNode))
+#define FREE_SACN_SOURCE_RB_NODE(ptr) free(ptr)
+#elif SACN_SOURCE_ENABLED
```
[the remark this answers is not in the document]
ChristianReese: private\common.h line 68

`#define SACN_SOURCE_ENABLED                                                                              \
  ((!SACN_DYNAMIC_MEM && (SACN_SOURCE_MAX_SOURCES > 0) && (SACN_SOURCE_MAX_UNIVERSES_PER_SOURCE > 0)) || \
   SACN_DYNAMIC_MEM)`

src/sacn/source.c
ChristianReese: The only function that uses this, sacn_source_create, documents the possible return codes that can be returned, and for what reason. I believe this line was my effort to translate internal error codes into a code that fits with the documentation. I do similar error-code-translation elsewhere as well. However I should probably do this translation in the actual API function and just return the code directly here.

src/sacn/source.c
ChristianReese: This function is pretty much process_sources + locking. It could either be a name that indicates that, or I could get rid of this function and just do the locking within the two functions that use it.

src/sacn/source.c
ChristianReese: If shutting_down is false then the thread will keep going even if there are no more thread-based sources (see keep_running_thread). The only way to get the thread to stop is to enable shutting_down, which is only set to true from sacn_source_deinit. That will automatically trigger termination for all thread-based sources as well, so at that point the thread keeps running until num_thread_based_sources inevitably falls to 0. So I think it already does what you want it to do.

src/sacn/source.c
ChristianReese: Not really, I'll go ahead and remove them.

src/sacn/source.c
ChristianReese: Ah, darn, didn't know that. Yeah, only things I can think of are either moving to arrays, or changing it so that it makes a separate local rbtree of things it wants to delete, and then goes through and deletes them after the iteration. I'll keep thinking about which option is best.

ChristianReese: Also good catch on the implementation not being fleshed out. I must have missed that when refactoring.

src/sacn/sockets.c
```diff
@@ -555,16 +639,77 @@ etcpal_error_t sacn_validate_netint_config(SacnMcastInterface* netints, size_t n
...
+        result = kEtcPalErrNoMem;
+#endif
+
+      if (result == kEtcPalErrOk)
+      {
+        for (size_t read_index = 0u, write_index = 0u; read_index < num_app_netints; ++read_index)
+        {
+          if (app_netints[read_index].status == kEtcPalErrOk)
+          {
+            memcpy(*internal_netints + write_index, &app_netints[read_index].iface, sizeof(EtcPalMcastNetintId));
```
samkearney, line 674: Unless I'm missing something, you can just do:
```
internal_netints[write_index] = app_netints[read_index].iface;
```

src/sacn/private/sockets.h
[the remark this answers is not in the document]
samkearney: OK, how about this then - create a struct type called SacnInternalNetintArray or similar which is contained by both the relevant source and receiver structs and contains the array (or array pointer in dynamic mem mode) and size. Then pass a pointer to that struct type to this function.

src/sacn/source.c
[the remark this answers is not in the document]
samkearney: I think it fits in the mem module if you think of it as containing the code necessary to implement allocation of the sACN structures, which includes their "constructors"

src/sacn/source.c
samkearney: Most applications won't implement logic based on the error code that's returned beyond checking for Ok vs not Ok, and will only use the differences in the other error codes for logging. So it's not completely critical for the documentation to capture every possible error code that will be returned from an API function, it should just make a "best effort" attempt. In RDMnet, I've written some hand-wavy statements like this in documentation:

https://github.com/ETCLabs/RDMnet/blob/5ba7466a05b98de864a0ff66ecf654ee57125970/src/rdmnet/broker/broker_api.cpp#L51

```
/// @return Other codes translated from system error codes are possible.
```

src/sacn/source.c
samkearney: Oh I see. The logic is so that you make sure you get the required Stream_Terminated packets out on shutdown. Gotcha

src/sacn/private/sockets.h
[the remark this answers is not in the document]
ChristianReese: Good idea, I'll go with that.

src/sacn/source.c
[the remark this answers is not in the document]
ChristianReese: Good point. I'll go ahead and pull this into add_sacn_source then.

src/sacn/source.c
ChristianReese: Good to know. I'll probably do the same thing in some places. Do you think there are still times when the error code should be "translated" between internal/external contexts? For instance, if I made a mistake in the source implementation and end up calling an internal function incorrectly, and it returns invalid - does it matter if I proceed to pass that error back to the application even if they fed in valid parameters? Probably not a huge difference, although it would make the logs slightly more clear and useful.

samkearney: Good question. In the situation you've mentioned, I think either option would result in an incorrect error code, because we don't have an error code for "there's a bug in the library".

If you translate the error code, it would return kEtcPalErrSys ("An internal library or system call error occurred"), which is not true.
If you don't, it will return Invalid which is also not true.
What will really happen is that the application user will debug, step down into the library, find the bug and file a bug report. So I don't really think it's a big deal.

ChristianReese: In that case I'll go through the code and simplify a lot of these translations out, and adjust documentation as needed.

src/sacn/source.c
```diff
@@ -152,66 +381,228 @@ void sacn_source_destroy(sacn_source_t handle)
...
+  etcpal_error_t result = kEtcPalErrOk;
+
+  // Verify module initialized.
+  if (!sacn_initialized())
+    result = kEtcPalErrNotInit;
+
+  // Check for invalid arguments.
+  if (result == kEtcPalErrOk)
+  {
+    if (handle == SACN_SOURCE_INVALID)
```
nwagner, line 404: Another if block to combine.

src/sacn/source.c
nwagner: That is a great discussion folks!  Good to keep in mind for the future.

src/sacn/source.c
nwagner: take_lock_and_process_sources?

src/sacn/source.c
nwagner: Arrays would seem to be the easier case, and perhaps we should start relying on them more than trees in general.  That invalidation of all iterators is really unfortuate.

src/sacn/source.c
nwagner: Perhaps pulling out the actual sending functionality as well.  I see several variations sending (unicast, multicast, IPv4, IPv6, IPv4 and IPv6)

src/sacn/mem.c
samkearney: @nwagner I may have made a wrong assumption here. I will get back to you

samkearney: @nwagner Yes, just confirmed my original comment was correct via a test. In the open-source red-black tree implementation we use, iterators save a traversal path from the root of the tree to the current node they are on. Removing an element can cause the tree to be rebalanced, which invalidates that path.

nwagner: Ok...  Can you update the rbtree documentation?  That's a nasty gotcha.
Adding doesn't rebalance the tree, then?

samkearney: Adding can definitely rebalance the tree, yes. I would expect any operation that modifies the tree to invalidate iterators.

src/sacn/source.c
ChristianReese: This has all been switched to arrays.

ChristianReese:
I made the bigger changes today, and have resolved the relevant threads. Still a bit left to go however, so stay tuned for more.

src/sacn/source.c
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/38cae99295d745bd5892951c99ccd3e0d377bef7 - saving further module refactors for when I start writing unit tests, if I see a further need.

src/sacn/private/common.h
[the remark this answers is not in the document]
ChristianReese: I improved the documentation for this in https://github.com/ETCLabs/sACN/pull/32/commits/85fd15bb8b82137598220a8bebf033f6fa61fdf5.

src/sacn/source.c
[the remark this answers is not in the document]
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/e73f667ab0f8dfd7376f21878fb96a3d6e6f438c.

src/sacn/source.c
```diff
@@ -152,66 +381,228 @@ void sacn_source_destroy(sacn_source_t handle)
...
+      {
+        SacnUnicastDestination* dest = add_sacn_unicast_dest(universe, &config->unicast_destinations[i]);
+        if (dest)
+        {
+          dest->terminating = false;
+          dest->num_terminations_sent = 0;
+        }
+        else
+        {
+          if (lookup_unicast_dest(handle, config->universe, &config->unicast_destinations[i], NULL) == kEtcPalErrOk)
```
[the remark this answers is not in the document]
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/9a88cc62e4c3fcb6011291027e0f8274c5237b9a.

src/sacn/source.c
ChristianReese: Improved documentation for this in https://github.com/ETCLabs/sACN/pull/32/commits/c634b9c92e56fdedecc51860acf8f997808cc002.

src/sacn/source.c
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/38cae99295d745bd5892951c99ccd3e0d377bef7.

src/sacn/mem.c
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/9a88cc62e4c3fcb6011291027e0f8274c5237b9a.

src/sacn/source.c
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/9a88cc62e4c3fcb6011291027e0f8274c5237b9a.

src/sacn/mem.c
ChristianReese: Mempools removed in https://github.com/ETCLabs/sACN/pull/32/commits/9a88cc62e4c3fcb6011291027e0f8274c5237b9a.

src/sacn/private/sockets.h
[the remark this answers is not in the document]
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/dcd53cdc3327dd8a8d4e6e8ef664245f344a88d4.

src/sacn/sockets.c
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/dcd53cdc3327dd8a8d4e6e8ef664245f344a88d4.

src/sacn/mem.c
```diff
@@ -35,6 +37,72 @@
...
+#if SACN_SOURCE_UNICAST_ENABLED
+#define ALLOC_SACN_UNICAST_DESTINATION() etcpal_mempool_alloc(sacnsource_unicast_dests)
+#define FREE_SACN_UNICAST_DESTINATION(ptr) etcpal_mempool_free(sacnsource_unicast_dests, ptr)
+#else
+#define ALLOC_SACN_UNICAST_DESTINATION() NULL
+#define FREE_SACN_UNICAST_DESTINATION(ptr)
+#endif
+#define ALLOC_SACN_SOURCE_RB_NODE() etcpal_mempool_alloc(sacnsource_rb_nodes)
+#define FREE_SACN_SOURCE_RB_NODE(ptr) etcpal_mempool_free(sacnsource_rb_nodes, ptr)
+#else
```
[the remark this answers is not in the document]
ChristianReese: This particular block of code no longer exists, but I did try to apply this in other areas where it could be helpful. See https://github.com/ETCLabs/sACN/pull/32/commits/a49e66e65071ffff22827c2004d033b07dadf149. I didn't comment the ones that had enough clarity already - meaning where there was no nesting and the corresponding if was close enough. But if you see more areas that should have a comment let me know. Resolving for now.

src/sacn/source.c
```diff
@@ -83,22 +189,101 @@ void sacn_source_universe_config_init(SacnSourceUniverseConfig* config)
...
+  etcpal_error_t result = kEtcPalErrOk;
+
+  // Verify module initialized.
+  if (!sacn_initialized())
+    result = kEtcPalErrNotInit;
+
+  // Check for invalid arguments.
+  if (result == kEtcPalErrOk)
+  {
+    if (!config)
```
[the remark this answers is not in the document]
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/4338ffdded0288b92cc19437c16825145d3966f0.

src/sacn/source.c
```diff
@@ -107,37 +292,81 @@ etcpal_error_t sacn_source_create(const SacnSourceConfig* config, sacn_source_t*
...
+  etcpal_error_t result = kEtcPalErrOk;
+
+  // Verify module initialized.
+  if (!sacn_initialized())
+    result = kEtcPalErrNotInit;
+
+  // Check for invalid arguments.
+  if (result == kEtcPalErrOk)
+  {
+    if (handle == SACN_SOURCE_INVALID)
```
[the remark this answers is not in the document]
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/4338ffdded0288b92cc19437c16825145d3966f0.

src/sacn/source.c
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/4338ffdded0288b92cc19437c16825145d3966f0.

src/sacn/source.c
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/4338ffdded0288b92cc19437c16825145d3966f0.

src/sacn/source.c
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/4338ffdded0288b92cc19437c16825145d3966f0.

src/sacn/source.c
```diff
@@ -83,22 +189,101 @@ void sacn_source_universe_config_init(SacnSourceUniverseConfig* config)
  */
 etcpal_error_t sacn_source_create(const SacnSourceConfig* config, sacn_source_t* handle)
 {
-  // If the Tick thread hasn't been started yet, start it if the config isn't manual.
+#if !SACN_SOURCE_ENABLED
+  return kEtcPalErrNotInit;
```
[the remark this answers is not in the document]
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/5bc6f38fd66d44978ad79db0f78b6f47cac8a4e6.

src/sacn/source.c
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/9546fa80cfb781cdedb46e4109a48d7ec78d2fc4.

src/sacn/source.c
ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/32/commits/fab065959ef9532213c8e2d6819f30fea1f929c5.

ChristianReese:
@samkearney @nwagner Alright, I believe I've implemented all of your feedback. You guys can review again now.

samkearney (approved):

nwagner (approved):
A lot of change here.  I know it's hard to make bulk changes easier for review.  Maybe next time we can break this out differently or into separate pull requests.

PR #33, opened 2021-01-29
Implemented Reset Networking for sACN Source

ChristianReese:
So far I've implemented Reset Networking for the source API (C & C++). I've also laid some of the groundwork for implementing it in the receiver API. But this PR will focus on the source implementation.

### DIFF
--- a/include/sacn/cpp/source.h
+++ b/include/sacn/cpp/source.h
@@ -130,6 +130,26 @@ public:
     bool IsValid() const;
   };
 
+  /**
+   * @ingroup sacn_source_cpp
+   * @brief A set of network interfaces for a particular universe.
+   */
+  struct UniverseNetintList
+  {
+    /** The source's handle. */
+    sacn_source_t handle;
+    /** The ID of the universe. */
+    uint16_t universe;
+
+    /** If !empty, this is the list of interfaces the application wants to use, and the status codes are filled in. If
+        empty, all available interfaces are tried. */
+    std::vector<SacnMcastInterface> netints;
+
+    /** Create an empty, invalid data structure by default. */
+    UniverseNetintList() = default;
+    UniverseNetintList(sacn_source_t source_handle, uint16_t universe_id);
+  };
+
   Source() = default;
   Source(const Source& other) = delete;
   Source& operator=(const Source& other) = delete;
@@ -164,14 +184,15 @@ public:
   void UpdateValuesAndForceSync(uint16_t universe, const uint8_t* new_values, size_t new_values_size,
                                 const uint8_t* new_priorities, size_t new_priorities_size);
 
-  etcpal::Error ResetNetworking(uint16_t universe);
-  etcpal::Error ResetNetworking(uint16_t universe, std::vector<SacnMcastInterface>& netints);
-
   std::vector<EtcPalMcastNetintId> GetNetworkInterfaces(uint16_t universe);
 
   constexpr Handle handle() const;
 
   static int ProcessManual();
+
+  static etcpal::Error ResetNetworking();
+  static etcpal::Error ResetNetworking(std::vector<SacnMcastInterface>& netints);
+  static etcpal::Error ResetNetworking(std::vector<UniverseNetintList>& netints);
 
 private:
   class TranslatedUniverseConfig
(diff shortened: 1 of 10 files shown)


src/sacn/source.c
```diff
@@ -1033,42 +1026,166 @@ int sacn_source_process_manual(void)
...
+
+      for (size_t j = 0; (result == kEtcPalErrOk) && (j < source->num_universes); ++j)
+      {
+        SacnSourceUniverse* universe = &source->universes[j];
+
+        universe->netints.num_netints = 0;
+        result = sacn_initialize_source_netints(&universe->netints, netints, num_netints);
+
+        for (size_t k = 0; (result == kEtcPalErrOk) && (k < universe->netints.num_netints); ++k)
+          result = add_to_source_netints(source, &universe->netints.netints[k]);
```
nwagner, line 1077: If you have 3 netints and the second one is bad, won't this drop the third netint as well?

src/sacn/source.c
```diff
@@ -1033,42 +1026,166 @@ int sacn_source_process_manual(void)
...
+
+      SacnSource* source;
+      SacnSourceUniverse* universe;
+      lookup_source_and_universe(netint_list->handle, netint_list->universe, &source, &universe);
+
+      universe->netints.num_netints = 0;
+      result = sacn_initialize_source_netints(&universe->netints, netint_list->netints, netint_list->num_netints);
+
+      for (size_t j = 0; (result == kEtcPalErrOk) && (j < universe->netints.num_netints); ++j)
+        result = add_to_source_netints(source, &universe->netints.netints[j]);
```
nwagner, line 1179: As before, we want all valid network interfaces, right?  Wouldn't this break early?

src/sacn/source.c
ChristianReese: I don't think so. My intent was for sacn_initialize_source_netints to initialize universe->netints with only the netints that have an OK status. So the second netint you mention that failed would not even make it into universe->netints. Only the first and third netints would be in universe->netints. So this loop would be k = 0 and k = 1 since universe->netints.num_netints would be 2.

src/sacn/source.c
ChristianReese: sacn_initialize_source_netints initializes universe->netints with only the valid netints. I don't see where this breaks early.

src/sacn/source.c
ChristianReese: Keeping in mind this was what I intended. It could very well still be breaking early, but I don't see where.

src/sacn/sockets.c
```diff
@@ -585,53 +540,165 @@ void sacn_send_unicast(sacn_ip_support_t ip_supported, const uint8_t* send_buf,
...
-      {
-        memcpy(&internal_netints->netints[write_index], &netints_to_use[read_index].iface, sizeof(EtcPalMcastNetintId));
-        ++write_index;
-      }
+      etcpal_ip_to_string(&netint->addr, addr_str);
     }
 
-    internal_netints->num_netints = num_valid_netints;
+    // Create a test send socket on each network interface. If it fails, we remove that interface from the respective
+    // set.
```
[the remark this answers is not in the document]
samkearney: @nwagner maybe not, but the inverse is more likely: If the socket can't be created now, it's unlikely that it will be created successfully in the future. This lets us avoid constantly trying to create sockets on invalid interfaces like Mac's utun, or interfaces associated with VM software, etc. every time a universe is created.

Btw, this is an adaptation of existing logic that was already in the library at 1.0.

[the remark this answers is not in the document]
ChristianReese: There definitely isn't a guarantee that a netint will have the same status, agreed there. However I'm not sure what we should do differently other than give the application a snapshot of where the netints are at currently. Maybe if one of the sockets stop working (we get an error from a socket call), we could trigger ResetNetworking internally?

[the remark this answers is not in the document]
nwagner: Ok, after thinking about your points I'm fine with keeping it as is.  If the sockets stop working, odds are that anything watching for network change will soon get a signal that the NIC is down anyway.

include/sacn/cpp/source.h
```diff
@@ -678,65 +710,107 @@ inline int Source::ProcessManual()
...
+ * @param[in, out] netint_lists Vector of lists of interfaces the application wants to use for each universe. Must not
+ * be empty. Must include all universes of all sources, and nothing more. The status codes are filled in whenever
+ * UniverseNetintList::netints is !empty.
+ * @return #kEtcPalErrOk: Networking reset successfully.
+ * @return #kEtcPalErrNoNetints: None of the network interfaces provided for a universe were usable by the library.
+ * @return #kEtcPalErrInvalid: Invalid parameter provided.
+ * @return #kEtcPalErrNotInit: Module not initialized.
+ * @return #kEtcPalErrSys: An internal library or system call error occurred.
+ */
+inline etcpal::Error Source::ResetNetworking(std::vector<UniverseNetintList>& netint_lists)
```
samkearney, line 796: I see why you added this function for completeness, but I don't know that anyone's use case will be this complex. We might want to leave it out until someone asks for it. It took me a bit to wrap my head around how to use it...

src/sacn/source.c
nwagner: Ah, I think the misunderstanding was on my side.  A closer look at add_to_source_netints() made me realize that the result check in the for loop was for a real error, not a bad interface.

src/sacn/source.c
nwagner: It does not.  The misunderstanding was on my side.

nwagner (approved):

include/sacn/cpp/source.h
nwagner: This is the only resetnetworking version that allows you to set different network interfaces for the different sources.  Since that was a request for source creation, we may need to keep this version, right?

src/sacn/source.c
ChristianReese: Yeah, there won't be an error unless all the interfaces are down.

include/sacn/cpp/source.h
samkearney: Yeah, we do. Never mind.

samkearney (approved):

include/sacn/cpp/source.h
ChristianReese: We resolved this offline.

PR #34, opened 2021-02-03
Implemented Reset Networking for sACN Receiver

ChristianReese:
Here's the Reset Networking implementation for the sACN Receiver.

### DIFF
--- a/src/sacn/receiver.c
+++ b/src/sacn/receiver.c
@@ -484,34 +484,68 @@ etcpal_error_t sacn_receiver_change_universe(sacn_receiver_t handle, uint16_t ne
 }
 
 /**
- * @brief Resets the underlying network sockets and packet receipt state for the sACN receiver..
+ * @brief Resets the underlying network sockets and packet receipt state for all sACN receivers.
  *
- * This is typically used when the application detects that the list of networking interfaces has changed.
+ * This is typically used when the application detects that the list of networking interfaces has changed, and wants
+ * every receiver to use the same network interfaces.
  *
- * After this call completes successfully, the receiver is in a sampling period for the universe and will provide
+ * After this call completes successfully, every receiver is in a sampling period for their universes and will provide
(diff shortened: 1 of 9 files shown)


src/sacn/receiver.c
```diff
@@ -845,6 +973,24 @@ void remove_receiver_sockets(SacnReceiver* receiver, bool close_now)
...
 }
 
+/*
+ * Remove all receiver sockets, choosing whether to close them now or wait until the next thread cycle.
+ *
+ * The sACN lock should be locked before calling this.
+ *
+ * [in] close_now Whether to close the sockets now or wait until the next thread cycle.
+ */
+void remove_all_receiver_sockets(bool close_now)
```
samkearney, line 983: I know that the remove_receiver_sockets() function was there before, but I'm not a fan of boolean flag parameters for stuff like this, they make it hard to read what's happening at the call site of the function:
```c
remove_all_receiver_sockets(true); // What does true mean??
```

If it's not too high impact, consider changing to an enum:
```c
typedef enum
{
  kCloseSocketNow,
  kQueueSocketForClose
} socket_close_behavior_t;

void remove_all_receiver_sockets(socket_close_behavior_t close_behavior)
```

samkearney (approved):

src/sacn/receiver.c
ChristianReese: Would you say there are still cases when boolean flag parameters should be used? If so what are those cases?

samkearney: They are appropriate when the function name gives you enough information to infer the parameter's usage from context. e.g.:

```c
set_rdm_enabled(true);
```

ChristianReese: Makes sense - in all cases, readability at the call site is key, so if it isn't readable with a bool, switch to an enum. I'll go ahead and make this change. Thanks!

nwagner: Good catch & discussion!

nwagner (approved):

samkearney (approved):

PR #36, opened 2021-02-05
New source_state module + source API unit tests (more to come)

ChristianReese:
This is the beginning of my unit test development effort for the sACN source API. I decided to split off the internal code into a separate source_state module. I have not written any unit tests for source_state yet, but I plan to. This way I can test the process_sources function and the more granular operations around it. The unit tests in this PR are testing the source API's interaction with the source_state module and other internal modules (so the main "_ Works" tests). Unit tests for testing validation and error cases are on my TODO list. Unit tests for source_state are also on my TODO list. So think of this PR as a "Part 1" for source unit tests.

I want to get your impressions on if I'm going in the right direction with this. And of course any other feedback you can provide.


tests/unit/api/c/source/test_source.cpp
```diff
@@ -35,6 +40,46 @@
 #define TestSource TestSourceStatic
 #endif
 
+#define NUM_TEST_NETINTS 3
+#define VERIFY_LOCKING(function_call)                                  \
```
nwagner, line 44: I like these!

nwagner (approved):
It feels like you're on the right track.  I like the refactoring of the source into source_state.  I think I need to learn more about fakes & custom_fakes.  This seems quite useful.

samkearney (approved):
I would definitely consider this the right direction. I like the separation into the API and source_state module.

PR #37, opened 2021-02-09
Unit tests for source_state module (part 1)

ChristianReese:
I've implemented some (not all) of the unit tests for the source_state module. In addition to general feedback, I want to get your thoughts on if I'm being too granular/detailed with my unit tests, or if I'm testing the right things at the right level of detail.

### DIFF
--- a/src/sacn/source_state.c
+++ b/src/sacn/source_state.c
@@ -477,8 +477,15 @@ int pack_universe_discovery_page(SacnSource* source, size_t* universe_index, uin
   // Update universe count, page, and last page PDU fields
   SET_UNIVERSE_COUNT(source->universe_discovery_send_buf, num_universes_packed);
   SET_PAGE(source->universe_discovery_send_buf, page_number);
-  SET_LAST_PAGE(source->universe_discovery_send_buf,
-                (uint8_t)(source->num_active_universes / SACN_UNIVERSE_DISCOVERY_MAX_UNIVERSES_PER_PAGE));
+  if (source->num_active_universes > 0)
+  {
+    SET_LAST_PAGE(source->universe_discovery_send_buf,
+                  (uint8_t)((source->num_active_universes - 1) / SACN_UNIVERSE_DISCOVERY_MAX_UNIVERSES_PER_PAGE));
+  }
+  else
+  {
+    SET_LAST_PAGE(source->universe_discovery_send_buf, 0);
+  }
 
   // Return number of universes packed
   return num_universes_packed;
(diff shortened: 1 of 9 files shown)


tests/unit/state/source/test_source_state.cpp
```diff
@@ -0,0 +1,382 @@
...
+  }
+
+  sacn_source_t AddSource(const SacnSourceConfig& config)
+  {
+    SacnSource* tmp = nullptr;
+    EXPECT_EQ(add_sacn_source(next_source_handle_++, &config, &tmp), kEtcPalErrOk);
+    return (next_source_handle_ - 1);
+  }
+
+  std::optional<SacnSource*> GetSource(sacn_source_t handle)
```
samkearney, line 101: I noticed a few things about this function:

It returns an optional, but when you use it, you always dereference the optional without checking it.
The optional value is a pointer, which already has a "null option" - a null pointer. Unless the null pointer has a different meaning than the null optional? Is there something I'm missing here?

samkearney (approved):

tests/unit/state/source/test_source_state.cpp
ChristianReese: Great catch - this can be simplified by having it just return a SacnSource*. I think I was trying to return an optional ref, and when I found out I couldn't, I switched to pointer without realizing how it could be simplified.

src/sacn/source_state.c
```diff
@@ -208,8 +209,8 @@ int process_sources(process_sources_behavior_t behavior)
 {
   int num_sources_tracked = 0;
 
-  // Iterate the sources backwards to allow for removals
-  for (size_t i = get_num_sources() - 1; i >= 0; --i)
+  // Iterate the sources backwards to allow for removals (i must not be unsigned)
+  for (int i = get_num_sources() - 1; i >= 0; --i)
```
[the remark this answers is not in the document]
ChristianReese: I'll have to take another pass and find more ways to switch from unsigned to signed. I was also trying to avoid conversion from signed to unsigned, but in the case of these loops I can make sure the value doesn't go negative, so the conversions should work if/when I need to make them.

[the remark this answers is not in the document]
ChristianReese: I'll make these changes in the next PR.

tests/unit/state/source/test_source_state.cpp
ChristianReese: This will be fixed in the next PR.

nwagner (approved):
I like where this is going.  The state module will be the trickiest to get right, so I think this level of unit testing is appropriate.

PR #39, opened 2021-02-11
Unit tests for source_state module part 2 + refactor size_t -> int

ChristianReese:
Putting out another PR extra early because I have a new blocker that will pull me away from this for a bit. More unit tests + changes requested in the last PR. More to come.


nwagner:
I think my first comment is a general one.  I'm seeing more (int) casts because of this, and I'm wondering if you should change more from size_t to int to help minimize the explicit casting.

ChristianReese:
Closing so I can do a bit more work on this, followed by another PR.

PR #40, opened 2021-02-16
Unit tests for source_state module part 2 version 2

ChristianReese:
Readjusted approach to size_t refactoring in response to feedback. More unit tests for source_state, but still not all of them. Other refactors/fixes as well.


src/sacn/source_state.c
```diff
@@ -319,8 +315,12 @@ void process_universe_null_pap_transmission(SacnSource* source, SacnSourceUniver
   {
     // Send 0x00 data & reset the keep-alive timer
     send_universe_multicast(source, universe, universe->null_send_buf);
-    send_universe_unicast(source, universe, universe->null_send_buf);
-    process_null_sent(universe);
+    send_universe_unicast(source, universe, universe->null_send_buf, kSkipTerminatingUnicastDests);
```
nwagner, line 318: I think we need to rename the functions and member variables, etc, that refer to startcode 0 as null.  At first glance, universe->null_send_buf implied to me that it was a null buffer.  process_universe_null_pap_transmission is also ambiguous with that mindset.  Even something as generic as "data" or "dmx" would avoid the issue.

RichardTea: I agree with Nick, "dmx" is clearer as to the purpose.

ChristianReese: Spoke offline with Nick and asked whether to use "dmx" or "level" and we ended up going with "level". Hopefully that will offer the same if not more clarity. Commit coming soon.

ChristianReese: Resolved in https://github.com/ETCLabs/sACN/pull/40/commits/f86be28fd19360833b579c79b751673ceb16b160

nwagner (approved):

PR #41, opened 2021-02-17
Source API treats terminating sources/universes/dests as removed

ChristianReese:
When I was writing unit tests I discovered an edge case where the application could technically modify (or at least access) the state of a source, universe, or unicast destination that is being removed (termination packets are being sent, so the underlying state is still there). This could be problematic and could open the door to bugs. This is my proposed solution. If a source/universe/dest is terminating but hasn't finished yet, the outward-facing API will still treat it as if it were removed and could not be found. I also adjusted the documentation a bit for this.


samkearney (approved):
I like this change.

nwagner (approved):
Good Catch.

PR #44, opened 2021-02-19
Unit tests for source_state module part 4

ChristianReese:
This is the remainder of the unit tests for the source_state.h functions. The remaining source unit testing work involves writing the unit tests for the source.h error cases, as well as for the C++ wrapper. These will probably end up being parts 5 and 6. Then I will most likely move on to the receiver unit tests, followed by completing test coverage for the internal modules.

### DIFF
--- a/tests/unit/state/source/test_source_state.cpp
+++ b/tests/unit/state/source/test_source_state.cpp
@@ -1805,3 +1805,13 @@ TEST_F(TestSourceState, DisablePapDataWorks)
   disable_pap_data(GetUniverse(source, universe));
   EXPECT_EQ(GetUniverse(source, universe)->has_pap_data, false);
 }
+
+TEST_F(TestSourceState, ClearSourceNetintsWorks)
+{
+  sacn_source_t source = AddSource(kTestSourceConfig);
+  AddUniverse(source, kTestUniverseConfig);
+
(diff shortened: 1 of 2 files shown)


tests/unit/state/source/test_source_state.cpp
```diff
@@ -1056,3 +1093,764 @@ TEST_F(TestSourceState, SourcesTerminateCorrectly)
...
+  SacnSourceUniverse* unicast_only_universe_state = nullptr;
+  lookup_source_and_universe(source, unicast_only_universe, &source_state, &unicast_only_universe_state);
+
+  update_levels_and_or_paps(source_state, unicast_only_universe_state, kTestBuffer, kTestBufferLength, nullptr, 0u,
+                            kDisableForceSync);
+  EXPECT_EQ(source_state->num_active_universes, 1u);
+}
+
+TEST_F(TestSourceState, IncrementSequenceNumberWorks)
+{
```
nwagner, line 1331: We need a version of this test that tests for proper rollover.

nwagner (approved):

tests/unit/state/source/test_source_state.cpp
ChristianReese: Great idea! I'll include that in the next PR.

PR #45, opened 2021-02-23
Use etcpal::OpaqueId in C++ Source

ChristianReese:
This PR involves updating EtcPal to the latest commit, fixing the resulting build failures, and finally switching Source::Handle to etcpal::OpaqueId.

### DIFF
--- a/include/sacn/cpp/source.h
+++ b/include/sacn/cpp/source.h
@@ -849,7 +849,7 @@ inline std::vector<EtcPalMcastNetintId> Source::GetNetworkInterfaces(uint16_t un
 /**
  * @brief Get the current handle to the underlying C sacn_source.
  *
- * @return The handle, which may or may not be valid.
+ * @return The handle, which will only be valid if the source has been successfully created using Startup().
  */
 inline constexpr Source::Handle Source::handle() const
 {
--- a/include/sacn/cpp/source.h
+++ b/include/sacn/cpp/source.h
@@ -294,7 +294,7 @@ inline etcpal::Error Source::Startup(const Settings& settings)
 inline void Source::Shutdown()
 {
   sacn_source_destroy(handle_.value());
-  handle_.SetValue(SACN_SOURCE_INVALID);
+  handle_.Clear();
 }
 
 /**
--- a/include/sacn/cpp/source.h
+++ b/include/sacn/cpp/source.h
@@ -39,6 +39,14 @@
 
 namespace sacn
 {
+
+namespace detail
+{
+class SourceHandleType
+{
+};
+};  // namespace detail
+
 /**
  * @ingroup sacn_source_cpp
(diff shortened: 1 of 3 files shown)


include/sacn/cpp/source.h
```diff
@@ -843,7 +849,7 @@ inline std::vector<EtcPalMcastNetintId> Source::GetNetworkInterfaces(uint16_t un
 /**
  * @brief Get the current handle to the underlying C sacn_source.
  *
- * @return The handle or Source::kInvalidHandle.
+ * @return The handle, which may or may not be valid.
```
samkearney, line 852: I think we should be more specific: "The handle, which will only be valid if the source has been successfully created using Startup()"

include/sacn/cpp/source.h
```diff
@@ -287,8 +293,8 @@ inline etcpal::Error Source::Startup(const Settings& settings)
  */
 inline void Source::Shutdown()
 {
-  sacn_source_destroy(handle_);
-  handle_ = kInvalidHandle;
+  sacn_source_destroy(handle_.value());
+  handle_.SetValue(SACN_SOURCE_INVALID);
```
samkearney, line 297: You can use handle_.Clear() as a shortcut here

samkearney (approved):

include/sacn/cpp/source.h
```diff
@@ -49,10 +50,12 @@ namespace sacn
 class Source
 {
 public:
+  struct HandleType
```
samkearney, line 53: Actually one more thing. I recommend moving this definition out of the Source class so it won't pop up on people's intellisense easily and confuse them. I would put it right above the Source class definition under the sacn namespace and in a detail namespace so it will be out of people's way.

Like so:
```cpp
namespace sacn
{

namespace detail
{
class SourceHandleType {};
};

class Source
{
public:
  using Handle = etcpal::OpaqueId<detail::SourceHandleType, sacn_source_t, SACN_SOURCE_INVALID>;

  // ...
};
```

include/sacn/cpp/source.h
ChristianReese: https://github.com/ETCLabs/sACN/pull/45/commits/6e80bb0110799e50953ba6a90aba3fe507716701

include/sacn/cpp/source.h
ChristianReese: https://github.com/ETCLabs/sACN/pull/45/commits/ebef5bebf5270959fa56c89c6cb1887fdd606c83

include/sacn/cpp/source.h
ChristianReese: https://github.com/ETCLabs/sACN/pull/45/commits/a3c9e020184c06f758fe78023a716e4ae028437f

nwagner (approved):

samkearney (approved):

PR #47, opened 2021-02-23
More source unit tests, mostly focusing on Source API error checking

ChristianReese:


tests/unit/api/c/source/test_source.cpp
```diff
@@ -40,7 +40,14 @@
 #define TestSource TestSourceStatic
 #endif
 
-#define NUM_TEST_NETINTS 3
+#define NUM_TEST_NETINTS 3u
```
samkearney, line 43: Always default to constexpr types instead of #define for constants.

```cpp
constexpr size_t kNumTestNetints = 3u;
// etc...
```

tests/unit/api/c/source/test_source.cpp
```diff
@@ -57,15 +72,35 @@
...
+  do                                                                             \
+  {                                                                              \
+    unsigned int previous_lock_count = sacn_lock_fake.call_count;                \
+    EXPECT_EQ(function_call, expected_return_value);                             \
+    EXPECT_EQ(sacn_lock_fake.call_count, previous_lock_count);                   \
+    EXPECT_EQ(sacn_lock_fake.call_count, sacn_unlock_fake.call_count);           \
+  } while (0)
 
 static const etcpal::Uuid kTestLocalCid = etcpal::Uuid::FromString("5103d586-44bf-46df-8c5a-e690f3dd6e22");
 static const std::string kTestLocalName = std::string("Test Source");
```
samkearney, line 85: These don't need to be std::strings, since you use c_str() to convert them back to const char* anyway.
```cpp
constexpr char kTestLocalName[] = "Test Source";
```

However, if you ever do need to define std::strings at file scope like this, just assign the string literal to them for more concise syntax:
```cpp
static const std::string kTestLocalName = "Test Source";
```

tests/unit/api/c/source/test_source.cpp
```diff
@@ -57,15 +72,35 @@
...
+static const EtcPalIpAddr kTestRemoteAddrs[NUM_TEST_ADDRS] = {
+    etcpal::IpAddr::FromString("10.101.1.1").get(), etcpal::IpAddr::FromString("10.101.1.2").get(),
+    etcpal::IpAddr::FromString("10.101.1.3").get(), etcpal::IpAddr::FromString("10.101.1.4").get()};
+static const EtcPalIpAddr kTestRemoteAddrsWithInvalid[NUM_TEST_ADDRS] = {
+    etcpal::IpAddr::FromString("10.101.1.1").get(), etcpal::IpAddr::FromString("10.101.1.2").get(),
+    etcpal::IpAddr().get(), etcpal::IpAddr::FromString("10.101.1.4").get()};
+static const sacn_source_t kTestHandle = 123;
+static const sacn_source_t kTestHandle2 = 456;
 static const uint16_t kTestUniverse = 456u;
+static const uint16_t kTestUniverse2 = 789u;
```
samkearney, line 100: Always default to constexpr
```cpp
constexpr uint16_t kTestUniverse2 = 789u;
```

tests/unit/api/c/source/test_source.cpp
```diff
@@ -77,8 +112,25 @@ static SacnMcastInterface kTestNetints[NUM_TEST_NETINTS] = {{{kEtcPalIpTypeV4, 1
...
+    {kTestHandle, kTestUniverse2, kTestNetints, NUM_TEST_NETINTS}};
+static const SacnSourceUniverseNetintList kTestInvalidNetintLists2[NUM_TEST_INVALID_NETINT_LISTS_2] = {
+    {kTestHandle, kTestUniverse, kTestNetints, NUM_TEST_NETINTS},
+    {kTestHandle2, kTestUniverse2, kTestNetints, NUM_TEST_NETINTS}};
+static const SacnSourceUniverseNetintList kTestInvalidNetintLists3[NUM_TEST_INVALID_NETINT_LISTS_3] = {
+    {kTestHandle, kTestUniverse, kTestNetints, NUM_TEST_NETINTS},
+    {kTestHandle, kTestUniverse2, kTestNetints, NUM_TEST_NETINTS},
+    {kTestHandle2, kTestUniverse, kTestNetints, NUM_TEST_NETINTS},
+    {kTestHandle2, kTestUniverse2, kTestNetints, NUM_TEST_NETINTS},
+    {kTestHandle2, kTestUniverse3, kTestNetints, NUM_TEST_NETINTS}};
```
samkearney, line 131: Is there a better way to organize these test values than 100 lines of constant definitions? The answer might well be no... but this is pretty hard to read and parse so it might be worth thinking about.

tests/unit/api/c/source/test_source.cpp
```diff
@@ -281,6 +694,19 @@ TEST_F(TestSource, SourceGetUniversesWorks)
   EXPECT_EQ(get_source_universes_fake.call_count, 1u);
 }
 
+TEST_F(TestSource, SourceGetUniversesHandlesNotFound)
+{
+  VERIFY_LOCKING(sacn_source_get_universes(kTestHandle, nullptr, 0u));
+  EXPECT_EQ(get_source_universes_fake.call_count, 0u);
+  SetUpSource(kTestHandle);
+  GetSource(kTestHandle)->terminating = true;
```
samkearney, line 702: Peeking into private state and manipulating it should always be a last resort when unit testing. Would it be the same test if you were to set the source to terminating by calling the API function to destroy it?

tests/unit/api/c/source/test_source.cpp
ChristianReese: https://github.com/ETCLabs/sACN/pull/47/commits/230a18a1140f41ba08a99dce914d2ae08bc49481

tests/unit/api/c/source/test_source.cpp
ChristianReese: https://github.com/ETCLabs/sACN/pull/47/commits/578a65f739891244091905dd93959c063d51a994

tests/unit/api/c/source/test_source.cpp
ChristianReese: https://github.com/ETCLabs/sACN/pull/47/commits/afa59c256bbb4d724ec85fb29f0748d8eed55ed9

tests/unit/api/c/source/test_source.cpp
ChristianReese: https://github.com/ETCLabs/sACN/pull/47/commits/2a7990d5464cdaf6a2e12a2595c12f309dcf3c67

samkearney (approved):
I like the use of vectors to simply the constants. Looks good. I don't have an answer for the private state thing, it's just something to keep in mind when writing tests - if a test is doable using just the observable behavior of the public interface (and underlying mocks), then that's the way it should be done. But sometimes it's not possible.

PR #49, opened 2021-02-26
Unit test source disabled in sacn_config.h

ChristianReese:
I'm introducing the idea of unit testing config settings. I'm starting by testing a sacn_config with the source disabled in static mode. Just creating a unit test project that disables the source API in the config has been incredibly useful, because initially the build actually failed because not all of the if blocks were in place for SACN_SOURCE_ENABLED. I also added a couple tests to make sure the return values are as expected with this configuration. Dynamic mode seems kind of out of place, but I figured it was worth checking that the API is enabled in this case, since the zeroed settings are ignored.


src/sacn/source_state.c
```diff
@@ -730,3 +733,10 @@ void remove_from_source_netints(SacnSource* source, const EtcPalMcastNetintId* i
       remove_sacn_source_netint(source, netint_index);
   }
 }
+
+#else  // SACN_SOURCE_ENABLED
+#ifdef _MSC_VER
+// Disable "nonstandard extension used" warning
```
samkearney, line 739: We shouldn't be ignoring "nonstandard extension used" warnings - that is the compiler telling you that you are doing something that might not be allowed by the C standard. I'm assuming it's complaining about having an effectively empty source file. If no functions in source_state.c are called when SACN_SOURCE_ENABLED=0, we can just leave it as-is without guarding it with #ifdefs; the linker will automatically remove the code from this file if it's not used by anything else.

tests/unit/CMakeLists.txt
```diff
@@ -2,13 +2,13 @@ include(GoogleTest)
 
 set(UNIT_TEST_DIR ${CMAKE_CURRENT_LIST_DIR})
 
-function(sacn_add_dynamic_test test_name)
+function(sacn_add_dynamic_test test_name config_dir)
```
samkearney, line 5: I think the original purpose of these functions was mainly to compile against different config file directories. So if we're adding the config directory as an argument, we might want to condense this to one function.

```cmake
function(sacn_add_dynamic_test test_name config_dir)
  set(target_name ${test_name})
  add_executable(${target_name} ${ARGN})

  target_compile_definitions(${target_name} PRIVATE SACN_HAVE_CONFIG_H)
  target_include_directories(${target_name} PRIVATE
    ${config_dir}
    ${SACN_INCLUDE}
    ${SACN_SRC}
  )
  target_link_libraries(${target_name} PRIVATE EtcPalMock gtest meekrosoft::fff)
  set_target_properties(${target_name} PROPERTIES CXX_STANDARD 17 FOLDER tests)
  gtest_discover_tests(${target_name})
endfunction()
```

And at the call site:
```cmake
sacn_add_test(test_merger_dynamic ${SACN_TEST}/configs/dynamic ${TEST_MERGER_SOURCES})
sacn_add_test(test_merger_static ${SACN_TEST}/configs/static ${TEST_MERGER_SOURCES})
```

tests/unit/config/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+# sACN config unit tests
+
+add_subdirectory(source_disabled)
```
samkearney, line 3: Following from my CMake test comment earlier, I think these config directories should be placed next to the other ones in the tree:
```
|- tests/
|-- unit/
|-- configs/
|--- dynamic/
|--- static/
|--- source_disabled_dynamic/
|--- source_disabled_static/
```

samkearney (commented):
Good thinking. This and other tests like it are going to be good to have in place to make sure we're not breaking compilations with various configurations when we modify the library in the future.

src/sacn/mem.c
```diff
@@ -112,7 +112,9 @@ typedef struct ToEraseBuf
 
 /**************************** Private variables ******************************/
 
+#if SACN_SOURCE_ENABLED
```
nwagner, line 115: Would it be worth breaking this file out into source_mem.c, etc?  I'm not particularly a fan of lots of #ifdefs sprinkled in a codebase.

tests/unit/CMakeLists.txt
nwagner: Personally, I liked the more specialized functions and not needing a config_dir passed in every time.

nwagner (commented):
I agree, a good idea.  It's really frustrating to start using "working" code and having it fail to build because your config is different from the usual.

src/sacn/common.c
```diff
@@ -74,9 +74,13 @@ etcpal_error_t sacn_init(const EtcPalLogParams* log_params)
...
+#if SACN_SOURCE_ENABLED
     bool source_state_initted = false;
+#endif
     bool receiver_initted = false;
+#if SACN_SOURCE_ENABLED
     bool source_initted = false;
+#endif
     bool merger_initted = false;
     bool merge_receiver_initted = false;
     bool source_detector_initted = false;
```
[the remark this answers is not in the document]
ChristianReese: https://jira.etcconnect.com/browse/SACN-61 has been created, so I'll resolve this, but let me know if you'd like this prioritized.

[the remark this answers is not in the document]
samkearney: Sounds good, we can hit it in the next sprint.

src/sacn/private/common.h
```diff
@@ -377,6 +377,8 @@ typedef struct SacnRecvThreadContext
  * Types used by the sACN Source module
  *****************************************************************************/
 
+#if SACN_SOURCE_ENABLED
```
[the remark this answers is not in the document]
ChristianReese: https://github.com/ETCLabs/sACN/pull/49/commits/6755f35cff4856a60753c813fbe8e7ede409303f https://github.com/ETCLabs/sACN/pull/49/commits/8b33cfb551932af29a0e42251b81f42b2817afe9 https://github.com/ETCLabs/sACN/pull/49/commits/aa8b0328a3d56468ffe5122d5b359229022ee887

src/sacn/source_state.c
ChristianReese: https://github.com/ETCLabs/sACN/pull/49/commits/6755f35cff4856a60753c813fbe8e7ede409303f https://github.com/ETCLabs/sACN/pull/49/commits/8b33cfb551932af29a0e42251b81f42b2817afe9 https://github.com/ETCLabs/sACN/pull/49/commits/aa8b0328a3d56468ffe5122d5b359229022ee887

ChristianReese: I'm assuming the linker will also remove them if the only functions that call them are also unused functions, and so forth.

src/sacn/mem.c
ChristianReese: https://github.com/ETCLabs/sACN/pull/49/commits/6755f35cff4856a60753c813fbe8e7ede409303f https://github.com/ETCLabs/sACN/pull/49/commits/8b33cfb551932af29a0e42251b81f42b2817afe9 https://github.com/ETCLabs/sACN/pull/49/commits/aa8b0328a3d56468ffe5122d5b359229022ee887

ChristianReese: I think there's slightly less of them in mem.c now, and they're much more isolated to the functions that reference the nonexistent variables. Let me know how it looks. I'm on the fence on source_mem.c since it wouldn't reduce the number of ifs.

tests/unit/config/CMakeLists.txt
ChristianReese: https://github.com/ETCLabs/sACN/pull/49/commits/ea1b6c5288b8eae72c7140acab09f0ebd4b4630e

tests/unit/CMakeLists.txt
ChristianReese: https://github.com/ETCLabs/sACN/pull/49/commits/e0484a21780d13efd1f9f88eda66567d14eb2557 https://github.com/ETCLabs/sACN/pull/49/commits/b941230b489640ed80e2a95bba8cd582266d18da

nwagner: So why are we passing in the config path on every call to sacn_add_[dynamic|static]_test?

ChristianReese: We no longer are. The only time the config path gets passed in is in the call to sacn_add_test, which is only called twice for the source disabled tests.

nwagner (approved):
When looking at "changes since last review" rather than "all changes", I was seeing the code changes I was looking for.  Thanks for clearing it up.

samkearney (approved):
I like the new CMake approach. Thanks

PR #51, opened 2021-03-03
Refactor sACN Receiver into internal modules

ChristianReese:
Similar to the source, the receiver has been split up into receiver, receiver_state, and mem. This will help with unit testing.

### DIFF
--- a/src/sacn/mem.c
+++ b/src/sacn/mem.c
@@ -37,6 +37,8 @@
 
 #define INITIAL_CAPACITY 8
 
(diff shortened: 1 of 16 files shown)


ChristianReese:
Oh wow, I just noticed that mem.c is pushing 2,000 lines of code... What do you guys think about splitting it up into receiver_mem and source_mem?

nwagner (approved):
Sure, it would make sense to split that out.

PR #53, opened 2021-03-08
Unit tests for receiver_state module part 2

ChristianReese:
While all the receiver_state.h functions had unit tests, the thread functionality did not have any unit tests for it. This is a significant part of the receiver_state code. Although this shouldn't typically be done, I brought an internal function (iterate_thread) into receiver_state.h purely so it could be accessed by unit tests. I did this because it was the best way I could think of to unit test the thread functionality, which covers the most important part of receiver_state. I will unit test this function very similarly to how I tested process_sources_and_lock in source_state. This will be a smaller PR because I wanted to get feedback earlier on if this is the right direction.

### DIFF
--- a/src/sacn/receiver_state.c
+++ b/src/sacn/receiver_state.c
@@ -294,6 +294,40 @@ void remove_all_receiver_sockets(socket_close_behavior_t close_behavior)
     remove_receiver_sockets(receiver, close_behavior);
 }
 
+/*
+ * Called in a loop by each receiver thread to manage incoming data and receiver state.
+ */
+void iterate_thread(SacnRecvThreadContext* context, EtcPalTimer* periodic_timer)
+{
+  if (sacn_lock())
+  {
+    sacn_add_pending_sockets(context);
+    sacn_cleanup_dead_sockets(context);
+    sacn_unlock();
+  }
+
+  SacnReadResult read_result;
+  etcpal_error_t read_res = sacn_read(context, &read_result);
+  if (read_res == kEtcPalErrOk)
+  {
+    handle_incoming(context->thread_id, read_result.data, read_result.data_len, &read_result.from_addr);
+  }
+  else if (read_res != kEtcPalErrTimedOut)
+  {
+    if (read_res != kEtcPalErrNoSockets)
+    {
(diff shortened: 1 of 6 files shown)


src/sacn/receiver_state.c
```diff
@@ -294,6 +291,46 @@ void remove_all_receiver_sockets(socket_close_behavior_t close_behavior)
     remove_receiver_sockets(receiver, close_behavior);
 }
 
+/*
+ * Called in a loop by each receiver thread to manage incoming data and receiver state.
+ */
+void iterate_thread(SacnRecvThreadContext* context)
```
nwagner, line 297: I think this function could have a more descriptive name.  It shouldn't be iterate_thread, rather it should be the functionality that the thread keeps calling. read_receiver_messages?

tests/unit/state/receiver/test_receiver_state.cpp
```diff
@@ -655,3 +718,70 @@ TEST_F(TestReceiverState, RemoveAllReceiverSocketsWorks)
...
+TEST_F(TestReceiverThread, AddsPendingSockets)
+{
+  sacn_add_pending_sockets_fake.custom_fake = [](SacnRecvThreadContext* recv_thread_context) {
+    EXPECT_EQ(recv_thread_context, get_recv_thread_context(0u));
+  };
+
+  for (unsigned int i = 1u; i <= 10u; ++i)
+  {
+    iterate_thread(get_recv_thread_context(0u));
+    sacn_add_pending_sockets_fake.call_count = i;
```
nwagner, line 731: I noticed a few of these tests iterate and don't seem to do much.  Are you just trying to see that the function is called 10 times, or was there more you were planning to add?

nwagner (approved):
Receiver_state is still an internal set of functions, so I don't think you need to worry about exposure.  As you say, this is critical logic to have some testing on.  Renaming the function will help, I think.

tests/unit/state/receiver/test_receiver_state.cpp
ChristianReese: The loop is just to demonstrate that the function is called once per iteration and isn't a one-off thing.

src/sacn/receiver_state.c
ChristianReese: I will change this in the next PR.

PR #57, opened 2021-03-12
Unit tests for receiver_state module part 6

ChristianReese:
A bit more receiver_state test code. The only thing I can think of left to do is to test the config with PAP disabled, which I'll get to on Monday.


samkearney (approved):
Nice, this is exactly the kinds of tests we're looking for right now. Critical parts of the code like the sequence number algorithm.

nwagner (approved):

PR #61, opened 2021-03-18
Merge receiver implementation part 1

ChristianReese:
I'm beginning the merge receiver implementation by adding the state struct and memory layer.


samkearney (approved):
LGTM. Loving the static constexpr declarations within the test function.

nwagner (approved):

PR #64, opened 2021-03-25
Merge receiver implementation part 4

ChristianReese:
This PR features the implementation of the universe_data callback and all the refactors resulting from it.

### DIFF
--- a/include/sacn/cpp/merge_receiver.h
+++ b/include/sacn/cpp/merge_receiver.h
@@ -100,13 +100,14 @@ public:
      * packets on the universe.
      *
      * @param[in] handle The merge receiver's handle.
+     * @param[in] universe The universe this merge receiver is monitoring.
      * @param[in] source_addr The network address from which the sACN packet originated.
      * @param[in] header The header data of the sACN packet.
      * @param[in] pdata Pointer to the data buffer. Size of the buffer is indicated by header->slot_count. This buffer
      *                  is owned by the library.
      */
-    virtual void HandleNonDmxData(Handle handle, const etcpal::SockAddr& source_addr, const SacnHeaderData& header,
-                                  const uint8_t* pdata) = 0;
+    virtual void HandleNonDmxData(Handle handle, uint16_t universe, const etcpal::SockAddr& source_addr,
+                                  const SacnHeaderData& header, const uint8_t* pdata) = 0;
 
     /**
      * @brief Notify that more than the configured maximum number of sources are currently sending on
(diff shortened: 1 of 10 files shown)


include/sacn/cpp/merge_receiver.h
```diff
@@ -100,14 +100,13 @@ class MergeReceiver
...
      * @param[in] handle The merge receiver's handle.
-     * @param[in] universe The universe this merge receiver is monitoring.
      * @param[in] source_addr The network address from which the sACN packet originated.
      * @param[in] header The header data of the sACN packet.
      * @param[in] pdata Pointer to the data buffer. Size of the buffer is indicated by header->slot_count. This buffer
      *                  is owned by the library.
      */
-    virtual void HandleNonDmxData(Handle handle, uint16_t universe, const etcpal::SockAddr& source_addr,
-                                  const SacnHeaderData& header, const uint8_t* pdata) = 0;
+    virtual void HandleNonDmxData(Handle handle, const etcpal::SockAddr& source_addr, const SacnHeaderData& header,
```
nwagner, line 108: I think I'd prefer to keep the universe as a parameter.  I agree that it is a bit redundant. but the other two callbacks return it and the application might be using the universe as an index to internal data instead of the handle.

src/sacn/merge_receiver.c
```diff
@@ -604,12 +603,114 @@ etcpal_error_t sacn_merge_receiver_get_source_cid(sacn_merge_receiver_t handle,
...
+    {
+      merger_handle = merge_receiver->merger_handle;
+      use_pap = merge_receiver->use_pap;
+
+      SacnMergeReceiverSource* source = NULL;
+      if (lookup_merge_receiver_source(merge_receiver, &header->cid, &source) == kEtcPalErrOk)
+      {
+        source_id = source->id;
+
+        if (source->pending && (header->start_code == 0x00))
```
nwagner, line 625: Let's add a comment here about the receiver always pushing 0x00 after 0xdd or timeout.  Otherwise my first impression was "what about pap"?

src/sacn/merge_receiver.c
```diff
@@ -604,12 +603,114 @@ etcpal_error_t sacn_merge_receiver_get_source_cid(sacn_merge_receiver_t handle,
...
+  {
+    SacnMergeReceiver* merge_receiver = NULL;
+    if (lookup_merge_receiver((sacn_merge_receiver_t)handle, &merge_receiver, NULL) == kEtcPalErrOk)
+      add_sacn_merge_receiver_source(merge_receiver, source_id, &header->cid, (header->start_code != 0x00));
+
+    sacn_unlock();
+  }
+
+  // Execute a new merge if needed.
+  bool new_merge = false;
```
nwagner, line 654: I think new_merge isn't quite the right name, because it's used to indicate that a new merge has taken place and a notification is needed.  I initially thought that the reader of the flag would execute the merge.  How about notify_merge?

nwagner (approved):
I'm assuming that you'll be adding unit tests next for universe_data()

include/sacn/cpp/merge_receiver.h
ChristianReese: https://github.com/ETCLabs/sACN/pull/64/commits/dbbd995afe1527a5cd2c810dc20deeea3f60961b

ChristianReese:
@nwagner I reverted the universe parameter commit. I'll resolve your other comments in the next PR.

src/sacn/merge_receiver.c
ChristianReese: > it's used to indicate that a new merge has taken place and a notification is needed
new_merge = true doesn't always mean a notification is needed. The only purpose is to pass on the news that a merge occurred, which factors into the decision whether to notify or not. But to avoid the confusion of if the merge was already done, I'll rename this to new_merge_occurred.

src/sacn/merge_receiver.c
ChristianReese: https://github.com/ETCLabs/sACN/commit/9a85950e3e4cd61be7c675e35a396f3da586b2f5

src/sacn/merge_receiver.c
ChristianReese: https://github.com/ETCLabs/sACN/commit/9ba306e80ec677c04fd7542bd466245671a5a55c

PR #72, opened 2025-09-10
Update CPM

ApolloFops:
The version of CPM that was being used depends on a deprecated CMake command, so this updates CPM so the builds will succeed.


ChristianReese:
We have reviewed and merged this change on our side, so I will be closing this PR. Thank you for this change!

PR #75, opened 2026-06-23
Fix bug in dmx_merger: remove PAP followed by add PAP can cause incorrect PAPs

samkearney:
Because sacn_dmx_merger_remove_pap() doesn't update the source's pap_count (it should reset it to the full slot count, as that's the result of spreading the universe priority across the full range), it can cause wrong behavior when PAP comes back with a short length, because trailing slot priorities past the new PAP length are not correctly zeroed.


ChristianReese (approved):
Trying out our new system that lets me merge the PR directly! Merging...

And thank you for the contribution @samkearney !

